PR #63, opened 2022-06-06
Migrate to testcontainers from docker compose

xnull:
I'd like to propose using [testcontainers](https://testcontainers.org) in unit tests. 
Testcontainers has multiple advantages in comparison to docker-compose:
 - a docker container being created in a unit test, no need for executing external commands.
 - it's more flexible in many ways (dynamic port bindings, creating containers using parameters from java source code, etc)
 - it manages the container's lifecycle in a unit test and cleans up all docker resources after the test finishes (disposable infrastructure approach)

This PR migrates PaymentProcessorTest class to testcontainers

### DIFF
--- a/pfi-mock-impl/build.gradle
+++ b/pfi-mock-impl/build.gradle
@@ -13,6 +13,8 @@ application {
     mainClass = 'io.tbd.tbdex.pfi_mock_impl.Main'
 }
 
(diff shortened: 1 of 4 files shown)


michaelneale:
thanks @xnull we have moved things to the dwn-js-sdk repo and ssi-sdk repos.

PR #69, opened 2022-06-20
Custom Message Deserializer

0e4b035a:
Issue https://github.com/TBD54566975/tbdex-protocol/issues/68


michaelneale:
thanks for this @0e4b035a - we have moved implementations outside (and in go/js)

PR #71, opened 2023-02-15
starting to clean this up

michaelneale:
Removing old implementations, fleshing out protocol and simplifying.


michaelneale:
@mistermoe @ALRubinger @angiejones I just want this to pass the test of "better than it was" as it better reflects the progress. I will be adding lots more of course.

mistermoe:
Will for sure take a look tonight!

mistermoe (approved):
yoyo @michaelneale thanks for rebooting this!

did a bit of housekeeping as well e.g. typos, markdown formatting, updating `CODEOWNERS`.

I Think we're in good shape for this PR. I've got some thoughts about the messages themselves but i'll create issues  w/ those thoughts and we can chat there so that this PR doesn't get stuck in review forever

PR #77, opened 2023-06-02
updated readme to add and update tbdex message types

jiyoontbd:
Updating the readme to include offering as a message type (see [this other issue](https://github.com/TBD54566975/tbdex-protocol/issues/75)), as well as update other message types that change as a result of offering being introduced.

### DIFF
--- a/protocol/README.md
+++ b/protocol/README.md
@@ -21,26 +21,31 @@ Every TBDex message contains the following fields:
 | `createdTime` | `long`        | Y              | The creation time of the message. Expressed as seconds since the Epoch                                                                |
 | `expiresTime` | `long`        | N              | The time after which this message should be considered as expired. Expressed as seconds since the Epoch                               |
 
+
+# ID for each message types
+The top-level TBDex `id` will serve as each message type's unique identifier. The message type is indicated with `type` field (i.e. `Offering`, `RFQ`, etc). `contextId` will serve as a way to identify a "thread" of messages sent back and forth between Alice and PFI.
+
 # Message Types
 
-The `body` of each message can be any of the following message types
+The `body` of each message can be any of the following message types.
 
 ## `Offering`
+> PFI -> Alice: "Here's what I can offer if you want to buy BTC with USD. Here are the constraints of my offer in terms of how much you can buy, what credentials I need from you, and what payment instruments you can use to pay me USD, and what payment instruments I can use to pay you BTC."
+
 | field            | data type | required | description                                                                                          |
 | ---------------- | --------- | -------- | ---------------------------------------------------------------------------------------------------- |
-| `id` | string    | Y        | A unique identifier for this offering.|
 | `description` | string    | Y        | Brief description of what is being offered.|
 | `pair` | string    | Y        | The currency pair being offered, in the format of `basecurrency_countercurrency`.|
 | `unitPrice` | string    | Y        | Price of 1 unit of base currency denominated in counter currency.|
 | `baseFee`   | string       | N        | Optional base fee associated with this offering, regardless of which Payment Instruments are used |
 | `min`   | string       | Y        | Minimum amount of counter currency that the counterparty (Alice) must submit in order to qualify for this offering.|
 | `max`   | string       | Y        | Maximum amount of counter currency that the counterparty (Alice) can submit in order to qualify for this offering.|
-| `presentationRequestJwt`   | string    | Y        |  PresentationRequest in JWT format which describes the credential needed to choose this offer.|
+| `presentationRequestJwt`   | string    | Y        |  PresentationRequest in JWT string format which describes the credential needed to choose this offer.|
 | `payinInstruments`   | list[PaymentInstrument]    | Y        |  A list of payment instruments the counterparty (Alice) can choose to send payment to the PFI from in order to qualify for this offering.|
 | `payoutInstruments`   | list[PaymentInstrument]    | Y        |  A list of payment instruments the counterparty (Alice) can choose to receive payment from the PFI in order to qualify for this offering.|
 
 ### Note on base and counter currency in `pair`
-There's an explicit directionality baked into the `pair` naming convention, which is `BaseCurrency_CounterCurrency`. Base Currency is the currency that the PFI is selling. Counter Currency is the currency that the PFI is willing to accept to sell the base currency (in other words, PFI is buying the Counter Currency). In trading terms, the side is always "SELL". 
+There's an explicit directionality baked into the `pair` naming convention, which is `BaseCurrency_CounterCurrency`. Base Currency is the currency that the PFI is **selling**. Counter Currency is the currency that the PFI is willing to accept to sell the base currency (in other words, PFI is **buying** the Counter Currency). In trading terms, the PFI's side is always "SELL". 
 
 ### `PaymentInstrument`
 | field            | data type | required | description                                                                                          |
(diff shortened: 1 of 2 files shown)


protocol/tbdex_schema_examples.js
```diff
@@ -0,0 +1,50 @@
...
+  "payout_instrument": {
+      "kind": "BTC_ADDRESS"
+  }
+}
+
+let quote = {
+  "rfqId": "1",
+  "quoteId": "2",
+  "expiryTime": "2023-04-14T12:12:12Z",
+  "totalFee": 1.00,
```
jiyoontbd, line 36: we're just assuming that Alice will be ok with fee being taken off from the original $10. what if she wanted to pay the fee on top of $10? what if the PFI wanted to offer that?

protocol/tbdex_schema_examples.js
```diff
@@ -0,0 +1,50 @@
...
+  "amount": 0.000383,
+  "paymentPresentationRequest": { }
```
jiyoontbd, line 38: includes both payin and payout VCs in 1 presentation request

protocol/tbdex_schema_examples.js
```diff
@@ -0,0 +1,50 @@
+let offering = {
```
jiyoontbd, line 1: overall comment - still undecided on if we need ids for each message type, and if so, what kinds of ids, or maybe a single transactionId across all message types?

protocol/tbdex_schema_examples.js
```diff
@@ -0,0 +1,50 @@
...
+  "pair": "BTC_USD",
+  "unitPrice": 27000.00,
+  "fee": 1.00,
+  "min": 10.00,
+  "max": 100.00,
+  "presentationRequest": { },
+  "payinInstruments": [{
```
jiyoontbd, line 8: is it possible that payment instruments (pay in or pay out) can affect the unit price? if that's the case, then should we have one payin/payout instrument combo per offer, instead of having arrays of all accepted payin/payout instruments per offer?

michaelneale:
I like it. One thing which is worth considering: the "pair" thing then puts rules on the sides of the currency exchange. Like they can't have "_" in them or be arbitrary strings, but also need a rule somewhere (implicit or explicit). For straight up currencies that is fine, but for goods exchange, could the be problematic? (one way is to have pair_left and pair_right and have it in the types).

protocol/tbdex_schema_examples.js
KendallWeihe: maybe illustrate that in the actual object here as an example

protocol/tbdex_schema_examples.js
KendallWeihe: `trace_id` I think is my vote... it's a concept which could be extended to other web5 protocols

protocol/tbdex_schema_examples.js
KendallWeihe: No I don't think that's ever the case. I do think different payment types offer different fee amounts, but a unit price is a universal number.

KendallWeihe:
> the "pair" thing then puts rules on the strings of the currency exchange

Oh man... this brings back bad memories from my days at Coin Metrics. We aggregated currency pairs across many exchanges, and the non-standard naming was always a headache. (https://coverage.coinmetrics.io/pairs) There's a name for this in the financial industry, but it's escaping me right now... something like a "security manifest" which is a canonical mapping of all possible naming conventions. Anyways...

It's a good point about exchanges with more than solely currencies, and I think we should account for the broader set of use cases. But idk maybe we can kick that can down the road?

protocol/README.md
```diff
@@ -25,147 +25,149 @@ Every TBDex message contains the following fields:
...
 | ---------------- | --------- | -------- | ---------------------------------------------------------------------------------------------------- |
-| `id` | string    | Y        | A unique identifier for this request.                                                                 |
-| `have` | string    | Y        | What you have.                                                                 |
-| `want` | string    | Y        | What you want.                                                                           |
-| `side`   | enum       | Y        | One of ["have", "want"] |
-| `size`   | int       | Y        | When side is "want", amount of want to spend on order. When "have", amount of have to spend on order.  |
-| `paymentTypes`   | JSON Object    | N        |  A normalized object describing the type of payments you need quotes for. When absent, the PFI is free to include any payment types.                 |
+| `kind` | enum    | Y        | Type of payment instrument (i.e. `DEBIT_CARD`, `BITCOIN_ADDRESS`)|
+| `fee` | string    | N        | Optional fee associated with using this kind of payment instrument.|
```
andresuribe87, line 48: Should fee be a separate object? That way it can evolve easily to different fee structures (or amount vs %).

protocol/README.md
```diff
@@ -25,147 +25,149 @@ Every TBDex message contains the following fields:
...
+| `pair` | string    | Y        | The currency pair being offered, in the format of `basecurrency_countercurrency`.|
+| `unitPrice` | string    | Y        | Price of 1 unit of base currency denominated in counter currency.|
+| `fee`   | string       | Y        | Base fee associated with this offering |
+| `min`   | string       | Y        | Minimum amount of counter currency that the counterparty (Alice) must submit in order to qualify for this offering.|
+| `max`   | string       | Y        | Maximum amount of counter currency that the counterparty (Alice) can submit in order to qualify for this offering.|
+| `presentationRequest`   | PresentationRequest    | Y        |  PresentationRequest which describes the credential needed to choose this offer.|
+| `payinInstruments`   | list[PaymentInstrument]    | Y        |  A list of payment instruments the counterparty (Alice) can choose to send payment to the PFI from in order to qualify for this offering.|
+| `payoutInstruments`   | list[PaymentInstrument]    | Y        |  A list of payment instruments the counterparty (Alice) can choose to receive payment from the PFI in order to qualify for this offering.|
```
andresuribe87, line 39: What would it mean for a payout instrument to have a fee associated to it? 

Perhaps PayoutInstrument should be a separate object?

protocol/README.md
```diff
@@ -25,147 +25,149 @@ Every TBDex message contains the following fields:
 
 The `body` of each message can be any of the following message types
 
-## `RequestForQuote`
-
+## `Offering`
+| field            | data type | required | description                                                                                          |
+| ---------------- | --------- | -------- | ---------------------------------------------------------------------------------------------------- |
+| `id` | string    | Y        | A unique identifier for this offering.|
+| `pair` | string    | Y        | The currency pair being offered, in the format of `basecurrency_countercurrency`.|
```
andresuribe87, line 32: Splitting this into an object, like what @michaelneale, seems like a good approach.

andresuribe87 (commented):
Awesome stuff!

protocol/README.md
```diff
@@ -25,147 +25,149 @@ Every TBDex message contains the following fields:
...
-  PFI->>Alice: PaymentReceipt
-  
+	actor PFI1
+  actor PFI2
+	actor PFIn
+	PFI1->>Alice: Offering
+	PFI2->>Alice: Offering
+	PFIn->>Alice: Offering
+  Alice->>PFI1: RequestForQuote (RFQ)
+  PFI1->>Alice: Quote
```
amika-sq, line 166: When Alice makes a RFQ, do we envision the returned quote coming back synchronously in the response? Or would Alice need to wait for the PFI to write the quote to their DWN?

protocol/README.md
```diff
@@ -25,147 +25,149 @@ Every TBDex message contains the following fields:
...
+  Alice->>PFI1: Order
+  PFI1->>Alice: OrderStatus
```
amika-sq, line 168: Similar question to the above here. Is this `OrderStatus` returned as a response, or something Alice needs to listen for on their DWN?

amika-sq, line 168: `OrderStatus` would represent the status at a particular moment in time. How do status updates get reported?

protocol/README.md
```diff
@@ -25,147 +25,149 @@ Every TBDex message contains the following fields:
...
 
-## `RequestForQuote`
-
+## `Offering`
+| field            | data type | required | description                                                                                          |
+| ---------------- | --------- | -------- | ---------------------------------------------------------------------------------------------------- |
+| `id` | string    | Y        | A unique identifier for this offering.|
+| `pair` | string    | Y        | The currency pair being offered, in the format of `basecurrency_countercurrency`.|
+| `unitPrice` | string    | Y        | Price of 1 unit of base currency denominated in counter currency.|
+| `fee`   | string       | Y        | Base fee associated with this offering |
```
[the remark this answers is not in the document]
jiyoontbd: the base fee is what i imagined everyone would pay, regardless of what payin/payout instrument they choose to do a transaction with the PFI. 

in the PaymentInstrument object, there's another fee (also probably going to become an object in the future too), for each payin/payout method.

the Quote includes a `totalFee` which should be the sum of base fee + fee associated with payment method that Alice chooses.

protocol/README.md
jiyoontbd: i think payout instruments can have fees associated with it. for example, a wire transfer could have a fee for moving money out of the PFI's account?

protocol/README.md
```diff
@@ -25,147 +25,149 @@ Every TBDex message contains the following fields:
...
 ```
-I have 100 AUD, and I want Cheese
+## `RequestForQuote`
+
+| field            | data type | required | description                                                                                          |
+| ---------------- | --------- | -------- | ---------------------------------------------------------------------------------------------------- |
+| `rfqId` | string    | Y        | A unique identifier for this Request for Quote. |
+| `pair` | string    | Y        | The currency pair being offered, in the format of `basecurrency_countercurrency`.|
+| `amount` | string    | Y        | Amount of counter currency you want to spend in order to receive base currency|
+| `verifiablePresentation` | VerifiablePresentation    | Y        | VerifiablePresentation that meets the specification per PresentationRequest in the Offering |
```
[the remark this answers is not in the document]
jiyoontbd: in the RFQ step, i was thinking that the `verifiablePresentation` would correspond to the "top-level presentationRequest" in the Offering, i.e. KYC VC (because PFI could require that you send in KYC VC no matter what payment method you choose).

during the Order step, the `paymentVerifiablePresentation` would correspond to the `paymentPresentationRequest` that was provided by the PFI, which would represent her payment VCs.

protocol/README.md
jiyoontbd: yeah, i was thinking fee would get more complex over time. i'll change it to be an object that just contains `flatFee` key for now.

protocol/README.md
```diff
@@ -25,147 +25,149 @@ Every TBDex message contains the following fields:
...
-| `costSize`      | int            | Y        | Amount of `costUnit` that is being requested for this offer.                            |
-| `paymentType`   | JSON Object    | Y        | A normalized object describing the type of payment acceptable for this offer.                        |
-| `presentationDefinitionRequest`     | JSON Object    | Y        | PresentationRequest that describes that credential requirements needed to accept this offer |
-
+| `rfqId`          | string         | Y        | The identifier of the RFQ request this quote is responding to.|
+| `quoteId`          | string         | Y        | The identifier of this quote.|
+| `expiryTime`     | datetime         | Y        | When this quote expires.|
+| `totalFee`     | string         | Y        | Total fee (base + paymentInstrument specific) included in quote|
+| `amount`     | string         | Y        | Amount of base currency that the PFI is willing to send|
+| `paymentPresentationRequest`     | PresentationRequest   | Y        | PresentationRequest that describes the payment instrument needed to execute this Quote (with payment kind indicated per the RFQ) |
```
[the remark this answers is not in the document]
jiyoontbd: so i think this would look a bit different in that this would smush 2 payment VC into 1 presentationRequest (payin and payout). 

this makes me think we don't actually need presentationRequest in the `Offering` object, since Alice is sending in `paymentVerifiablePresentation` against this `paymentPresentationRequest`. what do you think?

protocol/README.md
jiyoontbd: i would think the latter - Alice would wait for the PFI to write the Quote to Alice's DWN.

protocol/README.md
jiyoontbd: yep, something Alice needs to listen for on their DWN

jiyoontbd: that is a great question for folks thinking more about messaging at the protocol level - i feel like i'm remembering someone mentioning webhooks

i suppose in a web2 world, Alice would ping PFI for GET/ OrderStatus endpoint, passing in orderId... but i think that's no longer the case in the web5 / DWN world.

would appreciate thoughts from @mistermoe @KendallWeihe

andresuribe87: I would suggest that we think more as an event driven architecture. I would imagine Alice listening on messages that get written to her DWN

mistermoe:
should we add a `description` property to `Offering` that can be used to describe an offering?

mistermoe:
should payment instruments have `min` and `max`?

protocol/README.md
```diff
@@ -25,147 +25,148 @@ Every TBDex message contains the following fields:
...
+  "rfqId": "1",
+  "pair": "BTC_USD",
+  "amount": 10.00,
+  "verifiablePresentation": { /* Alice's KYC VC in Verifiable Presentation format */ },
+  "payinInstrument": {
+      "kind": "DEBIT_CARD"
+  },
+  "payoutInstrument": {
+      "kind": "BTC_ADDRESS"
+  }
```
amika-sq, line 94: The `kind` here is of type `BTC_ADDRESS`, but where would Alice provide their actual BTC address to the PFI? Would that be within this `payoutInstrument` object, or some other area?

protocol/README.md
jiyoontbd: so i'm going to actually punt on this and keep it as `pair` for now, with the caveat that this will definitely evolve (just like what we talked about with the `fee` key). i'd like to wrap up the conversation so we can get _something_ written down for initial projects, but we can definitely have this discussion again. also cc'ing @michaelneale since he first suggested this pair be split

protocol/README.md
jiyoontbd: so, during the `RFQ` step, Alice actually would not present her actual BTC address to the PFI. this is done during the `Order` step: 

```
{
    "orderId": 32432,
    "quoteId": 2,
    "paymentVerifiablePresentation": { /* Alice's Debit Card and Bitcoin VC in Verifiable Presentation format  */}
}
```

the `paymentVerifiablePresentation` is what's submitted to the PFI. it contains 2 x VCs, one for pay-in (her debit card) and another for pay-out (her actual BTC address). 

the general idea is that 
1. Alice says during the RFQ step "here's how i intend to pay", then
2. PFI's Quote says "ok, if you're gonna pay with those, here's how to shape your payment details whenever you're ready to send that in"
3. Alice's Order says "here's my payment details shaped how you want them"

does that make sense?

mistermoe:
> @mistermoe
> 
> > should payment instruments have min and max?
> 
> hmm... i think we can still make that work with min and max as top level fields.
> 
> for example, if a PFI didn't want Alice to pay with a credit card above $1000, he can include the CREDIT_CARD payinInstrument object in the offering that's for `min: 10.00, max: 999.99`...
> 
> but _not_ include it in a separate offering that has `min:1000.00, max: 5000.00`
> 
> i think that's easier to understand than having min and max for each paymentInstrument... also, it would only apply to payinInstruments, and not payoutInstruments, right?

`min` and `max` would be optional, similar to `presentationRequest`. 

🤔 so, separate offers entirely for different mins and maxes?

protocol/tbdex_schema_examples.js
jiyoontbd: done! presentationRequest would actually be in the format of a JWT strings, so i changed it and added a "realistic" presentation request jwt which you can decode using jwt.io to peek at the presentation definition inside.

mistermoe:
after a group chat we've decided to shelve the `min/max` per payment instrument for now

protocol/README.md
michaelneale: yeah it is async

michaelneale (approved):
I think this moves it along nicely!

protocol/tbdex_schema_examples.js
michaelneale: I think most things I have used - it just comes out of it. Never had a great experience though.

PR #78, opened 2023-06-13
Scaffold JS implementation

mistermoe:
# Overview

Scaffolded a typescript project under `implementations/js`. Goal is to include all type declarations and helper functions needed to construct tbdex messages. will publish it to npm once we have some meat on the bones. Also planning on including:
* json schemas for each tbdex message type
* dwn protocol definitions for both alice and pfi

--- 
added @jiyoontbd to `CODEOWNERS`


protocol/implementations/js/src/types.ts
```diff
@@ -0,0 +1,32 @@
+export namespace TbDex {
+  export type Offering = {
+    placeholder: string;
```
jiyoontbd, line 3: did you want to use all the fields we just changed per the readme changes in https://github.com/TBD54566975/tbdex-protocol/pull/77, or do that later?

mistermoe: down to go either way! could include them in this PR orrr could merge this to main and do a fast-follow PR with the 🥩

protocol/implementations/js/src/types.ts
```diff
@@ -0,0 +1,32 @@
...
+  export type MessageType<M extends keyof MessageTypes> = MessageTypes[M];
+  
+  export type Message<T extends keyof MessageTypes> = {
+    id: string;
+    contextId: string;
+    from: string;
+    to: string;
+    type: T;
+    body: MessageTypes[T];
+    dateCreated: string;
```
jiyoontbd, line 28: there's an `expiresTime` in the current readme - are we not using that field now? (tbh i was never 100% sure what that field would be for. the only time we use expiry is when we give out a Quote, and that would live in the `body` field anyway).

protocol/implementations/js/src/types.ts
jiyoontbd: let's do a fast follow PR just to keep PRs _atomic_ :p

jiyoontbd (approved):

protocol/implementations/js/src/types.ts
mistermoe: 💯 agreed. when i saw it i couldn't remember why i added it at the top-level because like you mentioned, expirations are really only needed for quotes

jiyoontbd: ok. i'll just write a smol issue to get rid of that field at the top tbdex message field level in the readme

Issue #81, opened 2023-06-15
implement json schemas for each tbDEX message type


mistermoe:
i think these should probably go in a `json-schemas` directory since json schema is language agnostic

PR #84, opened 2023-06-16
Implement json schemas

phoebe-lew:
- Add WIP schema definition
- Readme and example schema updates

Addresses #81

### DIFF
--- a/protocol/json-schemas/tbdex-message-schema.json
+++ b/protocol/json-schemas/tbdex-message-schema.json
@@ -202,27 +202,17 @@
             "$ref": "#/definitions/metadata"
         },
         {
-            "type": "object",
-            "properties": {
-                "type": {
-                    "enum": [
-                        "offering",
-                        "rfq",
-                        "quote",
-                        "order",
-                        "orderStatus"
-                    ]
-                }
-            }
-        },
-        {
             "oneOf": [
                 {
                     "type": "object",
                     "required": [
+                        "type",
                         "body"
                     ],
                     "properties": {
+                        "type": {
+                            "const": "offering"
+                        },
                         "body": {
                             "$ref": "#/definitions/offering"
                         }
@@ -231,9 +221,13 @@
                 {
                     "type": "object",
                     "required": [
+                        "type",
                         "body"
                     ],
                     "properties": {
+                        "type": {
+                            "const": "rfq"
+                        },
                         "body": {
                             "$ref": "#/definitions/rfq"
                         }
@@ -253,9 +247,13 @@
                 {
                     "type": "object",
                     "required": [
+                        "type",
                         "body"
                     ],
                     "properties": {
+                        "type": {
+                            "const": "order"
+                        },
                         "body": {
                             "$ref": "#/definitions/order"
                         }
@@ -264,9 +262,13 @@
                 {
                     "type": "object",
                     "required": [
+                        "type",
                         "body"
                     ],
                     "properties": {
+                        "type": {
+                            "const": "orderStatus"
+                        },
                         "body": {
                             "$ref": "#/definitions/orderStatus"
                         }
--- a/protocol/json-schemas/tbdex-message-schema.json
+++ b/protocol/json-schemas/tbdex-message-schema.json
@@ -17,6 +17,9 @@
                 },
                 "createdTime": {
                     "type": "number"
+                },
+                "expiryTime": {
+                    "type": "number"
                 }
             },
             "required": [
@@ -134,26 +137,25 @@
             "properties": {
                 "paymentVerifiablePresentationJwt": {
                     "type": "string"
-                },
-                "required": [
-                    "paymentVerifiablePresentationJwt"
-                ]
-            }
+                }
+            },
+            "required": [
+                "paymentVerifiablePresentationJwt"
+            ]
         },
         "orderStatus": {
             "type": "object",
             "properties": {
                 "orderStatus": {
-                    "type": {
-                        "enum": [
-                            "PENDING"
-                        ]
-                    }
-                },
-                "required": [
-                    "orderStatus"
-                ]
-            }
+                    "enum": [
+                        "PENDING",
+                        "COMPLETED"
+                    ]
+                }
+            },
+            "required": [
+                "orderStatus"
+            ]
         },
         "paymentInstrument": {
             "type": "object",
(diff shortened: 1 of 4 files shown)


protocol/tbdex_schema_examples.js
```diff
@@ -1,15 +1,15 @@
 let offering = {
   "description": "Buy BTC with USD!",
   "pair": "BTC_USD",
-  "unitPrice": 27000.00,
-  "baseFee": 1.00,
-  "min": 10.00,
-  "max": 1000.00,
+  "unitPrice": "27000.00",
```
KendallWeihe, line 4: nice! this looks great! 

only thing I'm curious about is why these types are strings instead of numbers?

andresuribe87: Precision matters in all financial applications. You don't want quantities to be truncated.

KendallWeihe: oh right, I see there was [discourse in Slack](https://sq-tbd.slack.com/archives/C05B492CW1H/p1686890655830929) about this

makes sense 👍

jiyoontbd (approved):

protocol/json-schemas/tbdex-message-schema.json
```diff
@@ -0,0 +1,276 @@
...
+                    "type": "string"
+                },
+                "from": {
+                    "type": "string"
+                },
+                "to": {
+                    "type": "string"
+                },
+                "createdTime": {
+                    "type": "number"
```
jiyoontbd, line 19: should this be
```
type: string
format: date-time
```
?

```

File: protocol/json-schemas/tbdex-message-schema.json [hunk 3]
```diff
@@ -0,0 +1,276 @@
...
+                "verifiablePresentationJwt",
+                "payinInstrument",
+                "payoutInstrument"
+            ]
+        },
+        "quote": {
+            "type": "object",
+            "properties": {
+                "expiryTime": {
+                    "type": "number"
```
jiyoontbd, line 109: same here :)

protocol/json-schemas/tbdex-message-schema.json
```diff
@@ -0,0 +1,276 @@
...
+                ]
+            }
+        },
+        "orderStatus": {
+            "type": "object",
+            "properties": {
+                "orderStatus": {
+                    "type": {
+                        "enum": [
+                            "PENDING"
```
jiyoontbd, line 149: we could add `INPROGRESS` and `COMPLETED` ?

protocol/json-schemas/tbdex-message-schema.json
```diff
@@ -0,0 +1,276 @@
+{
+    "definitions": {
+        "metadata": {
+            "type": "object",
+            "properties": {
```
jiyoontbd, line 5: missing `type` property? 
image

jiyoontbd: nvm i think i see how you did it below

```"properties": {
                "type": {
                    "enum": [
                        "offering",
                        "rfq",
                        "quote",
                        "order",
                        "orderStatus"
                    ]
                }
            }```

File: protocol/json-schemas/tbdex-message-schema.json [hunk 6]
```diff
@@ -0,0 +1,276 @@
...
+                "paymentInstructions"
+            ]
+        },
+        "order": {
+            "type": "object",
+            "properties": {
+                "paymentVerifiablePresentationJwt": {
+                    "type": "string"
+                },
+                "required": [
```
jiyoontbd, line 138: i think this required array needs to sit _outside_ of properties key?

protocol/json-schemas/tbdex-message-schema.json
```diff
@@ -0,0 +1,276 @@
...
+            "type": "object",
+            "properties": {
+                "orderStatus": {
+                    "type": {
+                        "enum": [
+                            "PENDING"
+                        ]
+                    }
+                },
+                "required": [
```
jiyoontbd, line 153: same here, i believe required array to sit outside of properties key so it's on the same level as orderStatus

protocol/json-schemas/tbdex-message-schema.json
jiyoontbd, line 149: i think this should change from

```
"orderStatus": {
    "type": {
        "enum": ["PENDING"]
     }
    ...
```

to

```
orderStatus": {
    "enum": ["PENDING"]
    ...
}
```

protocol/json-schemas/tbdex-message-schema.json
```diff
@@ -0,0 +1,276 @@
...
+                    "properties": {
+                        "body": {
+                            "$ref": "#/definitions/orderStatus"
+                        }
+                    }
+                }
+            ]
+        }
+    ]
+}
```
jiyoontbd, line 276: here's what chad said about `$schema`: 
image


i think including $schema would be good. i'm not really sure what we'd use for $id right now. maybe if this schema is published somewhere later, we can use the url for $id later?

protocol/json-schemas/tbdex-message-schema.json
phoebe-lew: Yep, cool, didn't know that was a thing!

phoebe-lew: Actually, these are longs/instants in the readme, which would be `number` instead of `datetime`. Following up in slack on whether they should be consistent types.

protocol/json-schemas/tbdex-message-schema.json
phoebe-lew: what's the difference between `PENDING` and `IN_PROGRESS`?

protocol/json-schemas/tbdex-message-schema.json
phoebe-lew: @jiyoontbd or @mistermoe the examples of $schemas i've seen are all URLs - is there one which makes sense for us?

michaelneale (approved):
some nice typo fixes too

protocol/json-schemas/tbdex-message-schema.json
jiyoontbd: nothing! they're the same in my mind. i was saying that we should probably come up with all possible order states. so it could be something along the lines of`PENDING`, `COMPLETED`, `ERROR` ?

phoebe-lew: added a couple but I think we'll have to revise as we go. For ex., can fail for pay in or pay out side, but do we want to distinguish that granularity here? and other questions like that.

PR #86, opened 2023-06-20
Implement DWN protocol

jiyoontbd:
Addresses #82 and some minor formatting changes in the readme


michaelneale (approved):
nice to see it in protocol form

protocol/implementations/js/src/protocolDefinitions.ts
```diff
@@ -0,0 +1,121 @@
...
+            '$actions': [
+                {
+                    'who': 'anyone',
+                    'can': 'write'
+                }
+            ],
+            // PFI is sending OUT quotes. no one should be writing Quotes to PFIs.
+            'Quote': {
+                '$actions': [],
+                // PFI who received an RFQ/Quote (i.e. Alice) can write Order to PFI's DWN
```
phoebe-lew, line 103: >PFI who received an RFQ/Quote (i.e. Alice)

Alice is a PFI?

protocol/implementations/js/src/protocolDefinitions.ts
```diff
@@ -0,0 +1,121 @@
...
+                // no one can read Order from PFI's DWN (except the PFI itself)
+                'Order': {
+                    '$actions': [
+                        {
+                            'who': 'recipient',
+                            'of': 'RFQ/Quote',
+                            'can': 'write'
+                        }
+                    ],
+                    // PFI is sending OUT OrderStatus. no one should be writing OrderStatus to PFIs.
```
phoebe-lew, line 113: > PFI is sending OUT OrderStatus.

How does Alice request an OrderStatus? or is it "push only" to alice?

phoebe-lew (approved):

protocol/implementations/js/src/protocolDefinitions.ts
michaelneale: I think that is a typo @jiyoontbd should be "Customer who received quote" as don't customers get quotes in response to RFQs? PFIs get RFQs and emit quotes in response.

jiyoontbd: > Alice is a PFI?

nope, i just can't write good like. apologies for the confusion!

what i meant by `PFI who received an RFQ/Quote (i.e. Alice)` was actually:

`only Alice, who received an RFQ/Quote, can write Order to PFI's DWN`

also, as an addendum, here i wrote the word `RFQ/Quote` to refer to `Quote` that Alice received from the PFI who received Alice's RFQ, in the `protocolPath` lingo, like how i refer to it in the `of: 'RFQ/Quote'`.

protocol/implementations/js/src/protocolDefinitions.ts
jiyoontbd: this is a great question! i would think that it's push only from PFI to Alice's DWN, and then Alice can read the orderStatus pushed to her DWN (like her "inbox"). right now i think of DWNs mailboxes, or maybe even messaging queues. i'd like @mistermoe @andresuribe87 to correct me if i'm wrong though.

andresuribe87: Yes, that's my same understanding!

One thing that's not currently supported is a way to fetch the `OrderStatus` on demand. Imagine that Alice has sent an `Order` to the PFI. After 2 minutes, she doesn't see any `OrderStatus` in her DWN. Is there a mechanism that she can use to "force" it?

diehuxx: One option is to mark Alice as the `recipient` of an `OrderStatus`, and add an `$action` rule allowing recipients to read. If Alice does not have a publicly addressable DWN, she can read the `OrderStatus` from PFI's DWN.

```
'OrderStatus': {
  '$actions': [{
    who: 'recipient',
    of: 'RFQ/Quote/Order/OrderStatus',
    can: 'read'
  }]
}
```

diehuxx (approved):
looks great!

protocol/implementations/js/src/protocolDefinitions.ts
KendallWeihe: Visual representation... yeah def **push** the `OrderStatus` IMO

image

protocol/implementations/js/src/protocolDefinitions.ts
```diff
@@ -0,0 +1,121 @@
+export const aliceProtocolDefinition = {
```
KendallWeihe, line 1: So we're going with two separate definitions as a mechanism of enforcing roles? Is that the idea?

protocol/implementations/js/src/protocolDefinitions.ts
diehuxx: @KendallWeihe Yes a PFI can write multiple `OrderStatus`s with the definition above.

protocol/implementations/js/src/protocolDefinitions.ts
jiyoontbd: yep, that's the idea! 

for this use case of two parties having distinct roles, and as dwn protocol definition is currently designed, we _must_ create two separate protocol definitions.

this is because if we had 1 protocol for both Alice and PFI, it's possible that Alice can write a Quote to PFI, or PFI can write an RFQ to Alice.

i do not think this is a great approach long term. i would like to just write one protocol and have it shared among multiple parties (that may play multiple distinct roles). this would require some more thinking on how dwn protocol definition is built in future versions.

protocol/implementations/js/src/protocolDefinitions.ts
michaelneale: @KendallWeihe how would a pull work? there would have the be a request for status message? 

Diagram is nice too - if it is helpful consider adding it to the readme of the repo perhaps?

jiyoontbd: added diagram to readme, with the edit of OrderStatus being a push from PFI to Alice :)

PR #87, opened 2023-06-20
Implement JS helper methods

### LINKED ISSUE #80: [js] implement helper methods to easily create tbDEX messages

phoebe-lew:
closes #80


protocol/implementations/js/src/builders.ts
```diff
@@ -0,0 +1,36 @@
+import { MessageTypes, TbDEXMessage } from './types.js'
+import {ulid} from 'ulidx'
+
+// maybe extract this somewhere?
+const ID_NAMESPACE = 'tbdex'
+
+function createId(messageType: string): string {
+  return `${ID_NAMESPACE}:${messageType}:${ulid()}`
```
phoebe-lew, line 8: temp implementation while the cbor/id requirements conversation continues: https://sq-tbd.slack.com/archives/C05B492CW1H/p1687222735830969

protocol/implementations/js/src/builders.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export function createMessage<Type extends keyof MessageTypes>(to: string, from: string, body: MessageTypes[Type], messageType: Type,): TbDEXMessage<Type> {
```
phoebe-lew, line 11: Is there a way to avoid having to pass in `messageType` again and to grab it from the generic?

protocol/implementations/js/tests/builders.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { expect } from 'chai'
+import { Offering } from '../src/types.js'
+import { createMessage } from '../src/builders.js'
+
+describe('Message builder', () => {
```
phoebe-lew, line 5: more tests needed obv

protocol/implementations/js/src/builders.ts
```diff
@@ -0,0 +1,36 @@
...
+  // TODO could validate that type is Offering or RFQ, since they start chains or are one off types
+  const id = createId(messageType)
+  return {
+    id          : id,
+    contextId   : id,
```
KendallWeihe, line 16: At what point is `contextId` different from `id`? Or will it always be different and you're just stubbing this in temporarily right now?

KendallWeihe: Just showcasing different syntax approach, one is not necessarily better than the other. I like exploring different syntax approaches for the intent of improving readability. Of course, readability, like beauty is in the eye of the beholder 😉 

```typescript
interface ICreateMessage {
  <Type extends keyof MessageTypes>(to: string, from: string, type: Type, body: MessageTypes[Type]): TbDEXMessage<Type>
}
export const createMessage: ICreateMessage = (to, from, type, body) => {
  const id = createId(type)
  return {
    id,
    contextId   : id,
    createdTime : new Date().toISOString(),
    to,
    from,
    type,
    body,
  }
}
```

KendallWeihe: btw if it weren't for the matter that `id` and `contextId` are the same thing, I would push it further to this... 

```typescript
export const createMessage: ICreateMessage = (to, from, type, body) => ({
  id: createId(type),
  contextId   : 'something else???',
  createdTime : new Date().toISOString(),
  to,
  from,
  type,
  body,
})
```

To me, this is very explicit in the intent... which is something like... _yeah the developer still have to pass in a majority of the properties, so it's not much different than just having them create the object on their own, but we're abstracting away the id & timestamping, which some devs may find verbose_

phoebe-lew: > At what point is contextId different from id?

It is always the same in this method! in the below function, where you pass in the last message, `contextId` will be different (pulled from the existing thread of messages).

phoebe-lew: @KendallWeihe bc I'm a TS newb, what do you think is the benefit of using `interface ICreateMessage`+`const createMessage: ICreateMessage` over a function type? Is this pattern still used when the interface only has one implementation?

mistermoe:
@phoebe-lew thoughts on this approach?

```typescript
import { MessageTypes, MessageType, TbDEXMessage, Metadata, PaymentInstrumentKind } from './types.js'

export type CreateMessageOpts<T extends keyof MessageTypes> = Partial<Metadata> & {
  from: string,
  to: string,
  messageType: T,
  body: MessageType<T>,
}

export function createMessage<T extends keyof MessageTypes>(opts: CreateMessageOpts<T>): TbDEXMessage<T> {
  console.log(opts)
}
```

benefits with this approach i think are:
1. can override all fields that we provide default values for
2. using an object as an argument makes it a bit more clear to the reader what fields they're setting (effectively faking named parrams in kotlin)
3. intellisense works nicely e.g.

when using the function, no explicit types are required but you still get intellisense for `body` based on the value chosen for `messageType`

https://github.com/TBD54566975/tbdex-protocol/assets/4887440/fd3e1c68-4cc1-4b07-886f-ad5011747d7d
 

using the variable also has intellisense without having to explicitly specify a type

image

protocol/implementations/js/src/builders.ts
mistermoe: i wish! all types are erased during compilation

protocol/implementations/js/src/builders.ts
KendallWeihe: `const` for the function (rather than `function`) is purely preferential... I like it because it reinforces the concept that everything in js is an object

> Is this pattern still used when the interface only has one implementation?

There is no pattern here, that I'm aware of. Again, purely subjective. I actually usually prefer inline typing, but in this case, there are so many types... idk to me it's easier to read by splitting it out

So yeah, beauty is in the eye of the beholder

phoebe-lew:
@KendallWeihe @mistermoe thoughts on this? I couldn't find a nice way of doing XOR types so resorted to full SO neckbeard. 

```
type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never }
type XOR<T, U> = (T | U) extends object ? (Without<T, U> & U) | (Without<U, T> & T) : T | U

type NewThreadOpts = {
  from: string,
  to: string,
}

type LastMessageOpts<T extends keyof MessageTypes> = {
  last: TbDEXMessage<T>
}

// should always require messageType, body
// should require XOR new thread opts/last messageopts. not both
export type CreateMessageOpts<T extends keyof MessageTypes, U extends keyof MessageTypes> = XOR<NewThreadOpts, LastMessageOpts<U>> & {
  messageType: T,
  body: MessageType<T>,
}

export function createMessage2<T extends keyof MessageTypes, U extends keyof MessageTypes>(opts: CreateMessageOpts<T, U>): TbDEXMessage<T> {
  const id = createId(opts.messageType)
  const contextId = opts.last?.contextId ?? id

  return {
    id          : id,
    contextId   : contextId ,
    createdTime : new Date().toISOString(),
    type        : opts.messageType        ,
    body        : opts.body,
    to          : opts.to ?? opts.last.to,
    from        : opts.from ?? opts.last.from,
  }
}
```

protocol/implementations/js/src/validator.ts
```diff
@@ -0,0 +1,34 @@
+import Ajv from 'ajv'
+import addFormats from 'ajv-formats'
+
+import tbdexMessage from '../../json-schemas/tbdex-message.schema.json' assert { type: 'json' }
```
phoebe-lew, line 4: @mistermoe this is causing the node tests to fail on build with:
`Import assertions are only supported when the '--module' option is set to 'esnext' or 'nodenext'.`, which is confusing because the root `tsconfig.json` does set module to esnext? 

Also doesn't happen ✨ on my machine ✨

KendallWeihe:
Nice! I think all of this looks fantastic!

With respect to this...

> @KendallWeihe @mistermoe thoughts on this? I couldn't find a nice way of doing XOR types so resorted to full SO neckbeard.

...I think this is great. Definitely prefer your solution over introducing a new dependency.

I'm glad you went w/ the `options` parameter approach... I almost recommended it, but I kinda don't love the syntax of passing objects as function params (I like primitives, even if there are a lot of them), except for at the root-level of applications (as in, up the call stack hierarchy), because objects are less explicit in their intent. That is, as a developer I can't _just_ look at a function signature, I have to look at a function signature _and_ a type. But, it's a common pattern, which we follow in a lot of our class constructors (ex. `new DwnProxy(options)`) and since we're in the world of Typescript it's more safe. Anyways... hehe 😅

protocol/implementations/js/src/builders.ts
```diff
@@ -0,0 +1,42 @@
+import { MessageType, MessageTypes, TbDEXMessage } from './types.js'
+import {ulid} from 'ulidx'
+
+type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never }
```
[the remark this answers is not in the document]
phoebe-lew: Ofc, good callout! I also added some comments explaining what they do

protocol/implementations/js/src/validator.ts
phoebe-lew: Looking at tsconfig again, does this include block:
```
  "include": [
    "./src",
  ],
  ```
potentially mean that it can't find `/protocol/implementations/json-schema` ? (although the error doesn't indicate that)

mistermoe: ah yup! it's being caused by `module` being set to `CommonJS` in `tsconfig.cjs.json` which was a flub on my part. just fixed and pushed.

PR #88, opened 2023-06-20
Generating typescript tbdex message types

jiyoontbd:
addresses #79

### DIFF
--- a/protocol/implementations/js/src/types.ts
+++ b/protocol/implementations/js/src/types.ts
@@ -9,6 +9,7 @@ export type TbDEXMessage = Metadata &
         body: Rfq
       }
     | {
+        type: "quote",
         body: Quote
       }
     | {
--- a/protocol/implementations/js/src/types.ts
+++ b/protocol/implementations/js/src/types.ts
@@ -1,26 +1,9 @@
-export type TbDEXMessage = Metadata &
-  (
-    | {
-        type: "offering"
-        body: Offering
-      }
-    | {
-        type: "rfq"
-        body: Rfq
-      }
-    | {
-        type: "quote",
-        body: Quote
-      }
-    | {
-        type: "order"
-        body: Order
-      }
-    | {
-        type: "orderStatus"
-        body: OrderStatus
-      }
-  )
+export type MessageType<M extends keyof MessageTypes> = MessageTypes[M];
+
+export type TbDEXMessage<T extends keyof MessageTypes> = Metadata & {
+  type: T;
+  body: MessageTypes[T];
+}
 
 export interface Metadata {
   id: string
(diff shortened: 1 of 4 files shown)


phoebe-lew (approved):
LFG

protocol/implementations/js/src/types.ts
```diff
@@ -1,32 +1,76 @@
...
-    contextId: string;
-    from: string;
-    to: string;
-    type: T;
-    body: MessageTypes[T];
-    dateCreated: string;
+export type TbDEXMessage = Metadata &
+  (
+    | {
+        type: "offering"
```
phoebe-lew, line 4: When we use this structure rather than the generics in the old version, do we still get the ability to check at a type level, that the message body "is of type Quote"?

protocol/implementations/js/src/types.ts
```diff
@@ -1,32 +1,76 @@
...
+}
+export interface PaymentInstruction {
+  link?: string
+  instruction?: string
+}
+export interface Order {
+  paymentVerifiablePresentationJwt: string
+}
+export interface OrderStatus {
+  orderStatus: "PENDING" | "COMPLETED" | "FAILED"
```
phoebe-lew, line 75: Is there any difference between this and using an `enum` type in TS?

protocol/implementations/js/src/types.ts
jiyoontbd: ooh yea i liked the generics stuff moe put in. it's just that i didn't get the generic stuff when i used a generator to create these types for me. it would be nice if the schema just did that for me.

i've fixed it to manually re-incorporate generics.

protocol/implementations/js/src/types.ts
jiyoontbd: mm maybe not but i like enums so i added that back in. i wonder if there was a way to say "make this output enums" in the schema...

protocol/implementations/js/src/types.ts
phoebe-lew: > used a generator to create these types for me

Oooh cool, so that goes from schema -> TS code?

protocol/implementations/js/src/types.ts
phoebe-lew: Ya i'm surprised the `enum` json schema type doesn't do that for you? >_<

KendallWeihe: my intuition is just build objects for enums, like this...

```typescript
export const Types = {
  Offering    : 'Offering',
  RFQ         : 'RFQ',
  Quote       : 'Quote',
  Order       : 'Order',
  OrderStatus : 'OrderStatus'
};
```

ts enums are terrible

protocol/implementations/js/src/types.ts
```diff
@@ -1,32 +1,81 @@
...
-    contextId: string;
-    from: string;
-    to: string;
-    type: T;
-    body: MessageTypes[T];
-    dateCreated: string;
+export type MessageType<M extends keyof MessageTypes> = MessageTypes[M];
+
+export type TbDEXMessage<T extends keyof MessageTypes> = Metadata & {
+  type: T;
```
KendallWeihe, line 4: I vote we go ahead and get in the habit of no semicolons 🙂

PR #91, opened 2023-06-24
Making Offering a tbdex resource

jiyoontbd:
borne out of discussion in slack about whether offering needs to be a tbdex message, or a new entity called `tbdex resource`: see summary of discussion in: https://hackmd.io/oBihor1TRcGIyMd5TMTrLQ


jiyoontbd:
> This makes sense, but I wasn't able to find the discussion (maybe there is not access)? It would be helpful to put the conclusions of the conversation in this PR, so anyone can see the history.

@andresuribe87 please see https://hackmd.io/oBihor1TRcGIyMd5TMTrLQ for a summary 😃

protocol/implementations/js/src/protocol-definitions.ts
```diff
@@ -39,17 +33,14 @@ export const aliceProtocolDefinition = {
...
-          OrderStatus: {
-            $actions: [
-              {
-                who : 'recipient',
-                of  : 'RFQ/Quote/Order',
-                can : 'write'
-              }
-            ]
-          }
+        OrderStatus: {
```
jiyoontbd, line 36: fyi @KendallWeihe we're removing `Order`, so now it's `RFQ -> Quote -> OrderStatus`. i did remove the empty $actions arrays, so i'd appreciate another set of :eyes: here to make sure it will still work

michaelneale (approved):
so happy to see Offering in here

PR #94, opened 2023-06-28
Compile validators

mistermoe:
This PR includes:

* Compiling message validators at compile/build time. 
    * Motivating factors behind doing this are:
        * to avoid dynamic code evaluation with Function constructor (used for schema compilation) - when it is prohibited by the browser page [Content Security Policy](https://ajv.js.org/security.html#content-security-policy). This applies to PWAs
        * prevents having to pack the json schemas themselves into the package. this gets weirdly tricky with `assert { type: 'json' }` and browser bundling gets even more 🤢 than it already is
    * Unintentional benefits: 
        *  reduces the browser bundle size - Ajv is not included in the bundle. bundle size isn't an issue right now. we basically got this for free
    * allegedly reduces the start-up time - the validation and compilation of schemas will happen during build time. didn't test the reduction myself, just something mentioned in the [Ajv docs](https://ajv.js.org/guide/managing-schemas.html#standalone-validation-code)
* Removal of polyfills from browser bundle since they aren't needed
* Exports dwn protocol definitions
* Splits schemas out into separate files. initially did this to chase down a bug. Ended up feeling like it was a bit easier to read
* modified `validateMessage` to do 2-phase validation. validates the outer message first and then validates the body based on the value of `payload.type`.
    * we have an issue open to return all errors instead of just the first one (#92). This is sort of in preparation for that. if `type` is not one of the expected message types, `body` would still be tested against each message type. the single schema approach understandably used `oneOf` which means each message type would have been attempted
    * Error messages end up being more readable (100% subjective)

### DIFF
--- a/protocol/implementations/js/.eslintrc.cjs
+++ b/protocol/implementations/js/.eslintrc.cjs
@@ -13,7 +13,8 @@ module.exports = {
   },
   ignorePatterns: [
     'dist',
-    'tests/compiled'
+    'tests/compiled',
+    'generated'
   ],
   rules: {
     'key-spacing': [
--- a/protocol/implementations/js/.gitignore
+++ b/protocol/implementations/js/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 dist
 tests/compiled
+generated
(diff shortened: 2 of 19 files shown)


phoebe-lew (approved):
Thanks for testing improvements + splitting up the schemas too. I reckon improves readability!

PR #95, opened 2023-06-28
Fixed build paths in package json

KendallWeihe:


mistermoe (approved):
Good catch! Was probably preventing the lib from being usable yeah?

KendallWeihe:
> Good catch! Was probably preventing the lib from being usable yeah?

indeed

PR #103, opened 2023-06-29
Alice's OrderStatus  fix

KendallWeihe:
The PFI is the author of the OrderStatus updates, so it should be the recipient of the `RFQ` not the `Quote` (Alice is the recipient of the quote)

### DIFF
--- a/protocol/implementations/js/src/protocol-definitions.ts
+++ b/protocol/implementations/js/src/protocol-definitions.ts
@@ -32,12 +32,12 @@ export const aliceProtocolDefinition = {
             can : 'write'
           }
         ],
-        // alice sends Orders, not receives them
+        // OrderStatus can be written to Alice's DWN by someone who wrote RFQ/Quote (i.e. PFI)
         OrderStatus: {
           $actions: [
             {
-              who : 'recipient',
-              of  : 'RFQ',
+              who : 'author',
+              of  : 'RFQ/Quote',
               can : 'write'
             }
           ]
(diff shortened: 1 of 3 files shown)


protocol/implementations/js/src/protocol-definitions.ts
```diff
@@ -37,7 +37,7 @@ export const aliceProtocolDefinition = {
           $actions: [
             {
               who : 'recipient',
-              of  : 'RFQ/Quote',
+              of  : 'RFQ',
```
jiyoontbd, line 40: request: could you get rid of the comment on L35? 🙏 i forgot to get rid of it when i got rid of Order

specifically this line: 
so i think here we want to say `OrderStatus can be written to Alice's DWN by someone who wrote RFQ/Quote (i.e. PFI)`

```js
$actions: [
            {
              who : 'author',
              of  : 'RFQ/Quote',
              can: 'write'
             }]
```

or `OrderStatus can be written to Alice's DWN by some who received an RFQ

```js
$actions: [
            {
              who : 'recipient',
              of  : 'RFQ',
              can: 'write'
             }]
```

looks like you went with the second option which makes sense to me, i think the original was wrong :)

jiyoontbd (approved):

protocol/implementations/js/src/protocol-definitions.ts
KendallWeihe: oh yeah... you know what, we should do this instead... notice how I'm using the `author` of the `RFQ/Quote`, which is your first comment proposal

```typescript
// OrderStatus can be written to Alice's DWN by someone who wrote RFQ/Quote (i.e. PFI)
OrderStatus: {
  $actions: [
    {
      who : 'author',
      of  : 'RFQ/Quote',
      can : 'write'
    }
  ]
}
```

done

PR #108, opened 2023-07-04
change type of `PaymentMethod.kind` to string

mistermoe:
Motivation behind this PR is two-fold:
* serializing a typescript enum returns a number with no context or meaning! e.g.

image

though, we can definitely do something like this if we want:

image

* hardcoding `kind`s limits the types of payment methods people can use with tbDEX. If someone wanted to use another payment method kind e.g. `MOMO` (mobile money), they'd have to come submit a PR to this repo.


phoebe-lew (approved):
y u like this typescript

phoebe-lew:
I do believe @KendallWeihe predicted this too!!

jiyoontbd:
note for future me, listen to kendall

jiyoontbd (approved):

PR #109, opened 2023-07-05
blehhhhh

phoebe-lew:
- Rename presentationRequest to presentationDefinition
- Yeet the protocol folder
- Yeet the implementations folder
- Fix GH workflow paths
- allow error response for rfq
- update protocol
- Fix interface
- Readme changes


KendallWeihe:
So, rather than introducing possible failure responses at every step in the TBDex message exchange, what if we introduced a singular `Status` message type, which would cover all of our use cases, including the (singular-in-nature) success stage of `COMPLETED`. Visual to help convey the idea...

image

The idea is, at any point, during the message exchange (notice, this doesn't include "resource's" like the `Offering`), either party can convey a `Status` message which would account for N-number of possible "failure" states, and then one singular `COMPLETED` state. In the same way unix error codes are `0` for success and some `non-0` for failure, where the code indicates the failure type.

I'll give some examples...

## RFQ --> Status = failure X

(I think this is the one you're solving for here)

This could come from the PFI, after Alice submits her RFQ, which would be the PFI saying, "Hey for reason X we can't give you a quote"

## Quote --> Status = failure Y

This could come from either party. Consider it originates from Alice, that could be Alice saying, "Hey I'm not going to accept that quote" (barring, Alice may not have to send a subsequent message whatsoever, she could just never progress it any further, but this would be a chance to provide a semantic reasoning). And then consider it originates from the PFI, that could be the PFI saying something like, "Hey you know that quote we gave you? Yeah that's no longer valid"

## Order --> Status = failure Z

This would come from the PFI, for some reason if the PFI couldn't complete the order.

Although, maybe this could _also_ come from Alice, and maybe that would be a chance for her to cancel the order... tho that would probably warrant a dedicated message type like `CancelOrder` b/c it's a request without guarantee... idk all ideas to prove the generalizable nature... 🤔 

## Order --> Status = `COMPLETED`

This is the one non-failure scenario, which the PFI would send to Alice once the order is successfully executed

PR #111, opened 2023-07-07
Changes

mistermoe:
# Changes
* use `PresentationDefinition` as value of `kycRequirements` instead of a JWT
    * there's really no good reason to return a JWT. we don't need cryptographic integrity here given that the entire message is signed by the PFI's DID 
* change `paymentPresentationDefinitionJwt` to `requiredPaymentDetails`
* use `JsonSchema` as value of `requiredPaymentDetails`
   * using PD's to collect payment details from Alice ends up being problematic given the expected flow of: "use VC that matches PD. if none present, render a form to collect necessary data, issue self-signed cred with input data. send self-signed VC." The 2nd time around, the VC generated as a byproduct of not having any will automatically be used and that may not be Alice's desired behavior (imagine that alice wants to use an entirely different btc address the 2nd time she uses DIDPay). Instead, we can just return a json schema. this json schema can be used to render a form every time. Alice's input data can then be sent in the RFQ as `paymentDetails`. The PFI can then validate the payment details provided against the `requiredPaymentDetails` json schema sent in the offering. 
* change `paymentVerifiablePresentationJwt` to `paymentDetails`
* use generic object as value of `paymentDetails`.
   * see `requiredPaymentDetails` explaination
* fix `Status` enum to serialize as string values instead of int values
* start fixing json schemas
* temporarily disable out validator tests
   * need to figure out how to validate offerings _and_ tbdex messages

### DIFF
--- a/js/src/types.ts
+++ b/js/src/types.ts
@@ -1,3 +1,9 @@
+import type { PresentationDefinitionV2 } from '@sphereon/pex-models'
+import type { Schema as JsonSchema } from 'ajv'
+
+export type { PresentationDefinitionV2, JsonSchema }
+
+
 export type ResourceType<R extends keyof ResourceTypes> = ResourceTypes[R]
 
 export type ResourceTypes = {
@@ -15,7 +21,7 @@ export interface Offering {
   baseFeeDollars?: string
   minDollars: string
   maxDollars: string
-  kycRequirements: string
+  kycRequirements: PresentationDefinitionV2
   payinMethods: PaymentMethod[]
   payoutMethods: PaymentMethod[]
   createdTime: string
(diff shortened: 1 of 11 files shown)


js/src/types.ts
```diff
@@ -15,15 +21,15 @@ export interface Offering {
   baseFeeDollars?: string
   minDollars: string
   maxDollars: string
-  kycRequirements: string
+  kycRequirements: PresentationDefinitionV2
```
phoebe-lew, line 24: does this lock all tbdex users into a pex depenency?

phoebe-lew (approved):

js/src/types.ts
phoebe-lew: *JS tbdex users

Issue #105, opened 2023-06-30
[js] Fee has dropped off Payin/outMethodResponse - add it back in


phoebe-lew:

SiennaSaito:
Looks like `PaymentMethod` in `types.js:33` has a `feeSubunits` property which doesn't seem to be reflected in either `./json-schemas/definitions.json` or `./json-schemas/tbdex-message.schema.json`.

Based off the readme `fee` was a dictionary but `feeSubunits` is a nullable String.
Is `feeSubunits` the replacement for `fee` or should they be seperate things? 🤔

SiennaSaito:
Also, yo! 😎 🌻

phoebe-lew:
Hi @SiennaSaito ! We just did a bigo refactor of the JS types this morning. the json schemas are sadly very out of touch with the types right now. Right now, the JS types should be considered the source of truth and we need to refactor the schemas (per #99, which is actually ALSO poorly named because the readme has not been updated to align with the JS changes)

SiennaSaito:
Okay sick! I'm running low on secret time today but I'll refresh at 4 PM, see if anything has changed and then make a list of things to change to match the JS types 🥸

SiennaSaito:
Alright - here's a small PR to update the readme with `feeSubunits` - let me know if this is on the right track.
https://github.com/TBD54566975/tbdex-protocol/pull/114

If that looks like it's on the right track then I'll put together a diff between the readme and `types.js`.

phoebe-lew:
Fee added back in in #111

PR #114, opened 2023-07-12
Update `fee` -> `feeSubunits` in README.md

SiennaSaito:
#105 

Updates references to `fee` in the readme to match `types.js`.
This is a small portion of the work needed to update the readme before the schema is also changed to match.


michaelneale (approved):

michaelneale:
thanks @SiennaSaito - wonder given that the types are now more rigorously defined if we could make the README automatically updated (or perhaps automated in some other way to not be repeated? if you have any ideas).

SiennaSaito:
Thanks @michaelneale!

If we're happy with adding the descriptions for each property to `types.js`, we could come up with a script that prints each type into a table and cats to stdout.

That'd get you 80% of the way there without having to worry about CI and templates etc.
If you wanted to hook it up with CI later then you could have CI update the readme automatically with each merge to main.

I'll spin up an issue so we can talk about it more ☀️

SiennaSaito:
https://github.com/TBD54566975/tbdex-protocol/issues/115

Issue #115, opened 2023-07-12
Readme tables aren't automated to match types.js


SiennaSaito:
# Context

Following up on @michaelneale's comment on #114 

`types.ts` appears to be the main source of truth for the tables in the readme which is currently manually updated.

If we take the descriptions laid out in the readme and inlined them in `types.ts` we could come up with a script that parses each type and automates generation of the readme table.

For example, types.js entries could look something like:

```
/** DESCRIPTIONS
  id: Unique identifier for this offering
  description: Brief description of what is being offered. 
  quoteUnitsPerBaseUnit: ???
  baseCurrency: Currency that the PFI is selling
  quoteCurrency: Currency that the PFI is accepting as payment for `baseCurrency`.
  kycRequirements: PresentationDefinition in JWT string format which describes the credential needed to choose this offer.
  payinMethods: A list of payment methods the counterparty (Alice) can choose to send payment to the PFI from in order to qualify for this offering.
  payoutMethods: A list of payment methods the counterparty (Alice) can choose to receive payment from the PFI in order to qualify for this offering.
  createdTime: The creation time of the resource. Expressed as ISO8601
*/
export type Offering = {
  id: string
  description: string
  quoteUnitsPerBaseUnit: string
  baseCurrency: CurrencyDetails
  quoteCurrency: CurrencyDetails
  kycRequirements: PresentationDefinitionV2
  payinMethods: PaymentMethod[]
  payoutMethods: PaymentMethod[]
  createdTime: string
}
```
or possibly:

```
export type Offering = {
  // Unique identifier for this offering
  id: string
  // Brief description of what is being offered. 
  description: string
  // ???
  quoteUnitsPerBaseUnit: string
  // Currency that the PFI is selling
  baseCurrency: CurrencyDetails
  // Currency that the PFI is accepting as payment for `baseCurrency`.
  quoteCurrency: CurrencyDetails
  // PresentationDefinition in JWT string format which describes the credential needed to choose this offer.
  kycRequirements: PresentationDefinitionV2
  // A list of payment methods the counterparty (Alice) can choose to send payment to the PFI from in order to qualify for this offering.
  payinMethods: PaymentMethod[]
  // A list of payment methods the counterparty (Alice) can choose to receive payment from the PFI in order to qualify for this offering.
  payoutMethods: PaymentMethod[]
  // The creation time of the resource. Expressed as ISO8601
  createdTime: string
}
```

[...]

PR #118, opened 2023-07-13
Automate drawing tables from types.js

SiennaSaito:
## Context
Following on from issue #115 and @michaelneale's comment on PR #114.

Automating the readme's table generation using `types.ts` as a source of truth.

I took a crack at it imperatively and its kinda gross - shoulda done something more sensible like:
* recursive descent parsing
* using the typescript AST
* maybe reflection?

#### BUT IT WORKS!

And I wrote it in bog standard ES6 so no need for any additional dependencies or plumbing. If you have node, it will run.

No worries if you don't want anything like this in the codebase at all - totally happy to get rejected outright on this.
It needs a *huge amount* of refactoring to be mergable, just raising this draft to work in the open.

## What works
The following currently seem to work when code is formatted idiomatically:
* types
* interfaces

The script produces pretty tables:

Screenshot 2023-07-14 at 12 18 02 am

It even prints out nullable types 😎

Screenshot 2023-07-14 at 12 47 29 am

In order for this to happen, types inside `types.ts` need to have comments attached to them:

Screenshot 2023-07-14 at 12 29 02 am

## What doesn't work
The following currently aren't supported:
* Single line exports 
* Enums
* Pulling nested types out into followup tables - right now nested types are written out as 'object' reflecting the type returned by `typeof(...)`.

Calculating nesting currently assumes only one brace per line so something like `{ a : { [String: Any] => Any } }` will break the script.

## Full output dump
Here's a complete dump of the current output. Github won't display the ANSI escape codes for colouring but they appear in the terminal and can be copied into any software that accepts rich text.

Note - I'd made some non-included changes to `types.ts`  prior to making this output so that there'd be some non missing descriptions inside the `Offering` table.

```
Table for ResourceTypes
-------------------------------------
| Name     | Type     | Description |
-------------------------------------
| offering | Offering | ???         |
-------------------------------------

Table for CurrencyDetails
----------------------------------------------
| Name         | Type          | Description |
----------------------------------------------
| currencyCode | string        | ???         |
| minSubunit   | string | null | ???         |
| maxSubunit   | string | null | ???         |
----------------------------------------------

Table for Offering
-----------------------------------------------------------------------------------------------------------------
| Name                  | Type                     | Description                                                |
-----------------------------------------------------------------------------------------------------------------
| id                    | string                   | The ID for the meow                                        |
| description           | string                   | The big meow is a big meow Note - do not do the big thing! |
| quoteUnitsPerBaseUnit | string                   | Okay so this is cool! Doubly so!                           |
| baseCurrency          | CurrencyDetails          | ???                                                        |
| quoteCurrency         | CurrencyDetails          | ???                                                        |
| kycRequirements       | PresentationDefinitionV2 | ???                                                        |
| payinMethods          | PaymentMethod[]          | ???                                                        |
| payoutMethods         | PaymentMethod[]          | ???                                                        |
| createdTime           | string                   | ???                                                        |
-----------------------------------------------------------------------------------------------------------------

Table for PaymentMethod
--------------------------------------------------------------------------------------------------------------
| Name                   | Type              | Description                                                   |
--------------------------------------------------------------------------------------------------------------
| kind                   | string            | valid kind strings: 'CASHAPP_PAY', 'APPLE_PAY', 'BTC_ADDRESS' |
| requiredPaymentDetails | JsonSchema | null | ???                                                           |
| feeSubunits            | string | null     | ???                                                           |
--------------------------------------------------------------------------------------------------------------

Table for MessageTypes
-------------------------------------------
| Name        | Type        | Description |
-------------------------------------------
| rfq         | Rfq         | ???         |
| quote       | Quote       | ???         |
| close       | Close       | ???         |
| orderStatus | OrderStatus | ???         |
-------------------------------------------

Table for MessageMetadata
---------------------------------------------
| Name        | Type          | Description |
---------------------------------------------
| id          | string        | ???         |
| threadId    | string        | ???         |
| parentId    | string | null | ???         |
| from        | string        | ???         |
| to          | string        | ???         |
| createdTime | string        | ???         |
---------------------------------------------

Table for TbDEXMessage<T extends keyof MessageTypes>  MessageMetadata &
----------------------------------------
| Name | Type            | Description |
----------------------------------------
| type | T               | ???         |
| body | MessageTypes[T] | ???         |
----------------------------------------

Table for Rfq
-------------------------------------------------------------
| Name                | Type                  | Description |
-------------------------------------------------------------
| offeringId          | string                | ???         |
| quoteAmountSubunits | string                | ???         |
| kycProof            | string                | ???         |
| payinMethod         | PaymentMethodResponse | ???         |
| payoutMethod        | PaymentMethodResponse | ???         |
-------------------------------------------------------------

Table for PaymentMethodResponse
------------------------------------------------
| Name           | Type          | Description |
------------------------------------------------
| kind           | string        | ???         |
| paymentDetails | object | null | ???         |
------------------------------------------------

Table for Quote
------------------------------------------------------------------
| Name                | Type                       | Description |
------------------------------------------------------------------
| expiryTime          | string                     | ???         |
| base                | object                     | ???         |
| quote               | object                     | ???         |
| paymentInstructions | PaymentInstructions | null | ???         |
------------------------------------------------------------------

Table for Close
----------------------------------------
| Name   | Type          | Description |
----------------------------------------
| reason | string | null | ???         |
----------------------------------------

Table for PaymentInstructions
----------------------------------------------------
| Name   | Type                      | Description |
----------------------------------------------------
| payin  | PaymentInstruction | null | ???         |
| payout | PaymentInstruction | null | ???         |
----------------------------------------------------

Table for PaymentInstruction
---------------------------------------------
| Name        | Type          | Description |
---------------------------------------------
| link        | string | null | ???         |
| instruction | string | null | ???         |
---------------------------------------------

Table for OrderStatus
--------------------------------------
| Name        | Type   | Description |
--------------------------------------
| orderStatus | Status | ???         |
--------------------------------------
```

### DIFF
--- a/scripts/generate-readme-tables.js
+++ b/scripts/generate-readme-tables.js
@@ -1,0 +1,609 @@
+const { readFile } = require('fs').promises
+
+const typesFilePath = __dirname + '/../js/src/types.ts';
+
+/////////////////////////////////////////////////
+///              🔥⭐️ Tokens ⭐️🔥             ///
+/////////////////////////////////////////////////
+
+class PropertyDeclaration {
+  constructor(name, type, description) {
+    this.name = name;
+    this.type = type;
+  }
+}
+
+class Comment {
+  constructor(text) {
+    this.text = text
+  }
+}
+
+class TableEntry {
+  constructor(declaration, comment) {
+    this.declaration = declaration;
+    this.comment = comment;
+  }
+}
+
+/////////////////////////////////////////////////
+///              🔥⭐️ Main ⭐️🔥               ///
+/////////////////////////////////////////////////
+
+readFile(typesFilePath, { encoding: 'utf8' })
+  .then(collectGroupedText)
+  .then(lexAndTokenize)
+  .then(reduceComments)
+  .then(constructTableEntries)
+  .then(drawTables)
+  .catch((error) => console.error(error));
+
+/////////////////////////////////////////////////
+///      🔥⭐️ Collecting Type Text ⭐️🔥       ///
+/////////////////////////////////////////////////
+
+// Input - A string containing the entire text of the file.
+// Output - A 2D array where each index contains the list of strings that make up a TS type definition
+// For Example, each index contains a list of strings separated by new lines that make up the following::
+//
+// export type Example {
+//   name: String
+//   description: String?
+// }
+//
+function collectGroupedText(rawTextData) {
+  let typeDefinitions = [];
+  let currentType = [];
+  let braceCounter = 0;
+  let lineNumber = 0; // For debugging
+
+  // doesn't currently support enums.
+  for (let line of rawTextData.split("\n")) {
+    lineNumber += 1;
+
+    // No inner structure - its just a single line export.
+    if (isASingleLineDefinition(line, braceCounter)) {
+      continue;
+    }
+
+    if (isAMultiLineDefinition(line, braceCounter)) {
+      braceCounter += 1;
+      currentType.push(line);
+      continue
+    }
+
+    if (braceCounter == 0) {
+      continue;
+    }
+
+    // A nested type.
+    if (isANestedTypeDefinition(line, braceCounter)) {
+      currentType.push(line);
+      braceCounter += 1;
+      continue;
+    }
+ 
+    // We've hit a closing brace
+    if (line.includes("}") && braceCounter > 0) {
+      braceCounter -= 1;
+      currentType.push(line);
+      if (braceCounter == 0 && line.includes('}')) {
+        typeDefinitions.push(currentType);
+        currentType = [];
+      }
+
+      continue;
+    }
+
+    if (braceCounter >= 0) {
+      currentType.push(line);
+      continue;
+    }
+  };
+  return typeDefinitions;
+}
+
+function isASingleLineDefinition(line, braceCounter) {
+  if (
+    (line.includes("export type") || line.includes('export interface'))
+    && (line.includes("}") || (!line.includes('{') && !line.includes('{'))) // One line definitions not supported.
+    && braceCounter == 0
+  ) {
+    return true;
+  }
+  return false;
+}
+
+function isAMultiLineDefinition(line, braceCounter) {
+  if (
+    (line.includes("export type") || line.includes('export interface'))
+    && !line.includes("}") 
+    && line.includes("{")
+    && braceCounter == 0
+  ) {
+    return true;
+  }
+  return false;
+}
+
+function isANestedTypeDefinition(line, braceCounter) {
+  // A nested type.
+  if (line.includes("{") && !line.includes("export type") && !line.includes("export interface") && braceCounter > 0) {
+    return true;
+  }
+  return false;
+}
+
+/////////////////////////////////////////////////
+///           🔥⭐️ Tokenizing ⭐️🔥            ///
+/////////////////////////////////////////////////
+
+class NestedTypeDefinition {
+
+  constructor() {
+    this.name = null;
+    this.isOptional = null;
+    this.originalIndex = null;
+  }
+
+  set(rawName, originalIndex) {
+    this.name = rawName.replace('?', '');
+    this.isOptional = isOptionalType(rawName);
+    this.originalIndex = originalIndex;
+    this.type = this.isOptional ? "object | null" : "object";
+  }
+
+  hasNoName() {
+    return this.name == null;
+  }
+
+  reset() {
+    this.name = null;
+    this.originalIndex = null;
+  }
+}
+
+// Processes the raw types into tokens - ie `PropertyDeclaration` and `Comment`
+// Replaces nested types with 'object'.
+// Input - A 2d array containing the line by line string representation of a Javascript type definition for each type inside a file
+// Output - A 2d list where each index contains a tuple of the type name and a list of either comments or property declarations.
+function lexAndTokenize(typeDefinitions) {
+  let result = [] // (outer name, [(inner name, type)])
+  let currentComment = ""
+
+  typeDefinitions.forEach((definition) => {
+    // Captures the title of the tile - eg `Offering` in `export type Offering`
+    let outerTypeName = extractOuterTypeNameFrom(definition);
+    let innerTypeStructure = innerTypeStructureFrom(definition);
+
+    // working out how to process nested types
+    let braceHeight = 0;
+    let i = 0;
+    let nestedTypeHolder = new NestedTypeDefinition();
+
+    // Iterate through the inner types
+    while(i < innerTypeStructure.length) {
+      line = innerTypeStructure[i];
+      // if you see an opening brace you've found an object
+      if (encounteringNestedType(line, nestedTypeHolder)) {
+        // record that we're operating on a nested type
+        // extract the name
+        nestedTypeHolder.set(line.split(":")[0], i);
+        innerTypeStructure.splice(i, 1);
+        braceHeight += 1;
+        continue;
+      }
+
+      // Delete the inner lines since we don't need them.
+      if (isInsideNestedType(line, braceHeight)) {
+        innerTypeStructure.splice(i, 1);
+        continue;
+      }
+
+      // we've currently looking at the lines inside a nested type
+      if (encounteringClosingBraceIn(line)) {
+        braceHeight -= 1;
+
+        // We're still inside a nested type.
+        if (braceHeight > 1) {
+          // throw away each line
+          innerTypeStructure.splice(i, 1);
+          continue;
+        }
+
+	// When we reach the final closing brace
+        if (braceHeight == 0) {
+          // Delete it.
+          innerTypeStructure.splice(i, 1);
+          // create a 'name: object' entry and insert it into the inner type structure
+          // thus representing the whole nested type as a complex object.
+          innerTypeStructure.splice(nestedTypeHolder.originalIndex, 0, "" + nestedTypeHolder.name + ": " + nestedTypeHolder.type);
+          nestedTypeHolder.reset();
+          i +=1;
+          continue;
+        }
+      }
+      i += 1;
+    }
+
+    // Captures the inner details as a `PropertyDeclaration` or a `Comment`
+    let innerDefinitions = innerTypeStructure.map((innerContents) => extractInnerTypeNameAndDefinitionFrom(innerContents));
+    result.push([outerTypeName, innerDefinitions])
+  });
+  return result; 
+}
+
+function isOptionalType(typeName) {
+  return typeName.includes('?') ? true : false;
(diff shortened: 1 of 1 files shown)


generate-readme-tables.js
SiennaSaito, line 1: Should put this in a scripts directory.

js/src/types.ts
```diff
@@ -98,6 +98,7 @@ export interface PaymentInstructions {
   payin?: PaymentInstruction
   payout?: PaymentInstruction
 }
+
```
quad, line 101: :trollface:

generate-readme-tables.js
```diff
@@ -0,0 +1,488 @@
...
+  TABLE_HEADER_NAME_TITLE: "Name",
+  TABLE_HEADER_TYPE_TITLE: "Type",
+  TABLE_HEADER_DESCRIPTION_TITLE: "Description",
+
+  BORDER: Foreground.BOLD + "| " + Styling.FINISHED,
+  TERMINAL_BORDER: Foreground.BOLD + "|" + Styling.FINISHED,
+//  STYLED_UNKNOWN_DESCRIPTION: Foreground.WHITE + Background.RED + "???" + Styling.FINISHED,
+//  STYLED_TABLE_HEADER_NAME_TEXT = "Name",
+//  STYLED_TABLE_HEADER_TYPE_TEXT = "Type",
+//  STYLED_TABLE_HEADER_DESCRIPTION_TEXT = "Description",
```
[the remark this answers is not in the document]
SiennaSaito: DUDE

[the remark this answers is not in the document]
SiennaSaito:
```
  ███▐       ██▌▒ █████████▒███████████▐█████████▐▓████▒  ▓██▒   ███▒    ███▒████████▌    
  ███▐       ██▌▌██▌▌          ▐██▒▒   ▐██▌▒      ▓█████▌ ▓██▐   ██▌▌    ██▌▌██▌▒  ▐███   
  ███▐       ██▌▌ ███████▒     ▐██▒    ▐███████   ▓██▐███▒▓██▐   ██▌▌    ██▌▌████████▐    
  ███▐       ██▌▌      ▐███▒   ▐██▒    ▐██▌▒      ▓██  ▐█████▐   ███▌    ██▌▌██▌▒         
  █████████▌▒██▌▌████████▐     ▐██▒    ▐█████████▐▐██    ████▐    ▓███████▌▒ ██▌▌         
```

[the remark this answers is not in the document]
SiennaSaito: STRTOK STROL

generate-readme-tables.js
```diff
@@ -0,0 +1,488 @@
...
+
+readFile('./js/src/types.ts', { encoding: 'utf8' })
+  .then((rawTextData) => collectTypeDefinitions(rawTextData))
+//  .then((collected) => collected.filter(x => shouldPrintForReadme(x[0]))) // Just focus on Offering for now
+  .then((typeDefinitions) => lexAndTokenize(typeDefinitions))
+  .then((tokens) => reduceComments(tokens))
+  .then((tokens) => constructTableEntries(tokens))
+  .then((tableEntries) => constructTableFrom(tableEntries))
+  .then((result) => console.log(result))
+  .catch((error) => console.error(error));
```
SiennaSaito, line 13: yeah probs don't need to lambda everything lol

generate-readme-tables.js
```diff
@@ -0,0 +1,488 @@
...
+
+function shouldPrintForReadme(outerTypeName) {
```
SiennaSaito, line 15: very big function 🥸

generate-readme-tables.js
```diff
@@ -0,0 +1,488 @@
...
+  return readmeResourceTypes.reduce((accumulation, next) => {
+      if (accumulation == true) { return true; }
+      return outerTypeName.includes(next) ? true : accumulation;
+    }, 
+    false
+  );
+}
+
+// Input - A string containing the entire text of the file.
+function collectTypeDefinitions(rawTextData) {
```
SiennaSaito, line 25: this is a very big function 🥸

generate-readme-tables.js
```diff
@@ -0,0 +1,488 @@
...
+    BRIGHT_CYAN: '\u001b[36;1m',
+
+    // Style Attributes
+    BOLD: '\033[1m',
+    UNDERLINE: '\033[4m',
+}
+
+const Constants = {
+  UNKNOWN_DESCRIPTION: "???",
+  TABLE_HEADER_NAME_TITLE: "Name",
```
SiennaSaito, line 446: These've gotta change to match the existing readme.
Also, probably good to hoist them to the top of the file lol

KendallWeihe:
Mad respect on this. Probs agree using the AST is the more robust approach; that is, the interfaces within the `typescript` npm package. But, parsing the raw text is a flex so kudos 💪. I don't mind merging this, tho I'm not sure how much usage it'll get. It seems to me, in a perfect world, we would automate this into the pipeline, which would actually write the results to a markdown file, as opposed to just console logging it out. It's a bit tricky putting it in the actual README, because that's a document which is highly edited, constantly mutating, and therefore it may make life chaotic putting the result in there. Not opposed, but would have to consider what it would look like to bridge that gap -- like, maybe we have a comment within the `README.md` which defines to the automation script where the contents should go. Else, IMO, a good solution would be to have a dedicated markdown document for these, and the README simply links to it. Have you considered packaging this up into an npm package? One which we could depend on here.

SiennaSaito:
Thanks for checking out the PR @KendallWeihe - and for making me feel like a pro for hand parsing 🤣

Totally getcha on the human element of docs - autogen always leaves something to be desired :(
Putting the comments inside `types.ts` defo adds a solid amount of visual noise. 

It might be tidier if the definitions lived above the outer type - maybe something like:
```
// @property - id - Unique identifier for this offering
// @property - description - Brief description of what is being offered.
// @property - quoteUnitsPerBaseUnit - Number of quote units on offer for one base currency unit (i.e 290000 USD for 1 BTC
export type Offering = {
id: string
description: string
quoteUnitsPerBaseUnit: string
}
```

But this adds a lot of extra text 🤔 

> It seems to me, in a perfect world, we would automate this into the pipeline, which would actually write the results to a markdown file, as opposed to just console logging it out. It's a bit tricky putting it in the actual README, because that's a document which is highly edited, constantly mutating, and therefore it may make life chaotic putting the result in there. Not opposed, but would have to consider what it would look like to bridge that gap -- like, maybe we have a comment within the README.md which defines to the automation script where the contents should go

Yeah - the doc feels like a living breathing target so I figured a good first step would be to provide the table and let a human paste it in. That being said - execution wise, by the time we get to the end of the script we have a list containing a model of each entity inside the file, so it'd be easy to just pick the types we're interested in and have CI insert them into a markdown file at a specific commented location. 

Happy to explore that more if you think its worth it!

> Have you considered packaging this up into an npm package? One which we could depend on here.
That's a cool idea! I'm actually a Swift dev and am just doing JS for fun so I never considered this.
I'll take a look into it on Monday 😎

SiennaSaito:
> I don't mind merging this ... I like the human touch to documentation too much to try and automate it. But, I'm almost certainly in the smaller camp on this opinion, and using comments to bridge the gap to auto-doc-gen is certainly the standard in many places ❤️

@phoebe-lew, @jiyoontbd, @mistermoe, reckon this kind of script would save you enough time to be worth the extra clutter inside `types.ts`?

scripts/generate-readme-tables.js
```diff
@@ -0,0 +1,609 @@
...
+  UNKNOWN_DESCRIPTION: "???",
+  TABLE_HEADER_NAME_TITLE: "Field",
+  TABLE_HEADER_TYPE_TITLE: "Data Type",
+  TABLE_HEADER_DESCRIPTION_TITLE: "Description",
+
+  BORDER: Foreground.BOLD + "| " + Styling.FINISHED,
+  TERMINAL_BORDER: Foreground.BOLD + "|" + Styling.FINISHED,
+}
+
+// Wow you read down to the bottom of the file!
```
jiyoontbd, line 580: ty ty

jiyoontbd:
hi @SiennaSaito ! thanks so much for this PR. like kendall said, it's a lot of work to parse text from types file and then parse again into human readable tables, many kudos  ✨ 

a bit of history of types.ts you may not know - our original intention was to generate types.ts from json schema, which live [here](https://github.com/TBD54566975/tbdex-protocol/tree/main/json-schemas). the schema-to-types generation isn't perfect, and since we were trying to use tbdex-protocol for the our pfi-prototype, we have been primarily making changes inside types.ts, because it was faster 😅 

however, since json schema is our eventually... _intended_ source of truth, what are your thoughts on generating the readme tables from the schema, not types.ts? (nb: this will require that we first update all the json schemas to match current types.ts cuz they're a bit outdated) so we'd use json schemas to generate *both* the types.ts and the readme tables. 

one way of going this route would be useful is that we can provide `description` for each field in json schema, so there's no need for requiring comments inside types.ts file. another is that the json schema is broken up per each type (i.e. RFQ, Quote, etc etc), so hopefully there will be less "climb down the bracket ladder" work required.

would love to get your thoughts!

mistermoe:
Really love the idea! +1 to @jiyoontbd's point. 

a few thoughts:
* there's quite a bit of prior art around generating docs / tables using types + comments that could be useful here. it's all rooted in [`TSDoc`](https://tsdoc.org/). [API Extractor](https://api-extractor.com/pages/setup/generating_docs/) and [TypeDoc](https://typedoc.org/) are two examples of projects that consume `TSDoc` and output API documentation.  Both  provide plugin entrypoints that can be used to write plugins to output customized docs if there's nothing off-the-shelf that does what we need. Both projects produce ASTs and call plugin methods while iterating through the AST
* Agreed that the json schemas should be our source of truth in hopes that codegen works for other languages as well. In terms of pre-existing libs, it seems like people go in the other direction (concrete types -> json schemas). [`typescript-json-schema`](https://www.npmjs.com/package/typescript-json-schema) is the goto lib in typescript land. worth mentioning that this lib uses tsdoc symbols as a means to extract stuff like `description`. this will likely get superceded when decorators make their way into ECMAScript (in [TC39 stage 3](https://github.com/tc39/proposal-decorators) atm)

SiennaSaito:
Thanks for the awesome input @jiyoontbd + @mistermoe! 🙏🏼

I'm going to close this PR out as the JSON schema is the eventual intended source of truth and merging this PR would probably cause more work to be done down the track (migrating the descriptions again).

I'm going on a personal sabbatical for an awhile - if the readme generation still needs to be done later down the track then I'll take it up after I've recharged. In the meanwhile I think this PR acts as a good set of notes in case anyone else wants to tackle it in the meanwhile! ☀️

PR #120, opened 2023-07-14
yeet the homerolled ids and replace with typeid-js

phoebe-lew:
For now, have only made the message id and offering ids strongly typed. parent/thread id are still strings - can discuss if this feels weird and we want parent/thread ids strongly typed as well. It's just slightly weird with parent id because the type depends on which message type you're responding to. 

Some other notes:
- offering ids are namespaced by offering
- message ids are namespaced by the _message type_
- message id continues to exist outside of `body`, but is moved out of the `metadata` type because it needs to know about the type.


js/src/types.ts
```diff
@@ -47,21 +48,21 @@ export type MessageTypes = {
 }
 
 export interface MessageMetadata {
-  id: string
-  threadId: string
+  threadId: string // technically this has to be an rfq id because that's how you start a thread
   parentId?: string
```
jiyoontbd, line 52: could i suggest that you also yeet this change this line to `parentId?: string | null` ? i dunno why but typescript gets upset when it's exported out as a node_module

image

jiyoontbd: nvm you've changed this bit
```const parentId = opts.last?.id.toString() ?? undefined```

File: js/src/types.ts [hunk 2]
```diff
@@ -47,21 +48,21 @@ export type MessageTypes = {
...
   from: string
   to: string
   createdTime: string
 }
 
 export type TbDEXMessage<T extends keyof MessageTypes> = MessageMetadata & {
+  id: TypeID<T>
```
jiyoontbd, line 59: ok so with this change, id is still a "top level" tbdex message field, it's just that it is an ID of type T using typeid instead of our own :)

jiyoontbd (approved):

js/tests/builders.spec.ts
```diff
@@ -32,7 +33,7 @@ describe('messages builders', () => {
   })
```
jiyoontbd, line 33: i think line above this one should now be
```    expect(actual.parentId).to.be.undefined
```

phoebe-lew: thank yewww!

js/src/types.ts
phoebe-lew: Correct! the message itself still looks the same and the interface for constructing a message is the same

PR #121, opened 2023-07-18
swapping from and to

jiyoontbd:
Swapping from and to when constructing message, in order to facilitate a conversation thread with correct from/to fields


corcillo (approved):

js/src/builders.ts
```diff
@@ -25,8 +25,8 @@ export function createMessage<T extends keyof MessageTypes, U extends keyof Mess
     id,
     threadId,
     parentId,
-    to          : opts.to ?? opts.last.to,
-    from        : opts.from ?? opts.last.from,
+    to          : opts.from ?? opts.last.from,
```
phoebe-lew, line 28: This reads pretty confusingly to someone without context, maybe add a comment explaining why they need to be swapped?

js/src/builders.ts
```diff
@@ -10,12 +10,23 @@ type LastMessageOpts<T extends keyof MessageTypes> = {
...
   type: T,
   body: MessageType<T>,
 }
 
-export function createMessage<T extends keyof MessageTypes, U extends keyof MessageTypes>(opts: CreateMessageOpts<T, U>): TbDEXMessage<T> {
+export type CreateMessageOpts2<T extends keyof MessageTypes, U extends keyof MessageTypes, X extends keyof MessageTypes> =
+XOR<NewThreadOpts, ReplyMessageOpts<U>> & {
+  type: T,
+  body: MessageType<T>,
+  last?: TbDEXMessage<X>
```
phoebe-lew, line 26: In which situations do you pass in `last` vs `reply`? And in which situations would you pass in _both_?

js/tests/builders.spec.ts
```diff
@@ -74,8 +74,8 @@ describe('messages builders', () => {
       body : quote
     })
 
-    expect(from).to.equal(rfqMessage.from)
-    expect(to).to.equal(rfqMessage.to)
+    expect(from).to.equal(rfqMessage.to)
```
phoebe-lew, line 77: would this make more sense if we renamed to `sender` and `recipient`? then the logic would read like:

`expect(quoteSender).to.equal(rfqMessage.recipient)` which seems somewhat more english-y

js/src/builders.ts
jiyoontbd: ok we have a smol matrix here

no last, no reply: initial message (i.e. RFQ) - i expect you to pass in to and from

yes last, no reply: replying to a thread where last message is from you (i.e. OrderStatus, Close?) - i expect you to pass in to and from

no last, yes reply:... i can't think of an example here.

yes last, yes reply: replying to a thread where last message is NOT from you (i.e. PFI sends Quote in response to having received RFQ) - i'll fill in the to and from for you

js/tests/builders.spec.ts
jiyoontbd: oh, i like that ^_^

jiyoontbd:
Yeah just close it for now please 🙏

PR #137, opened 2023-08-17
REST API spec

### LINKED ISSUE #124: add restful API spec
defined in: https://hackmd.io/rPXZD9iyQO6CCuF6guWOTA?view

mistermoe:
closes #124


phoebe-lew (approved):

rest-api/README.md
```diff
@@ -0,0 +1,321 @@
...
+- [References](#references)
+
+# Discoverability
+PFIs can become publicly discoverable by advertising their API endpoint as a [Service](https://www.w3.org/TR/did-core/#services) within their DID Document. In order to increase the likelihood of being discovered The `service` entry **SHOULD** include the following properties:
+
+| Property          | Value                                                         |
+| ----------------- | ------------------------------------------------------------- |
+| `id`              | see [DID-CORE spec](https://www.w3.org/TR/did-core/#services) |
+| `type`            | `PFI`                                                         |
+| `serviceEndpoint` | PFI's publicly addressable API endpoint                       |
```
michaelneale, line 72: ``
```suggestion
| `serviceEndpoint` | PFI's publicly addressable API endpoint or DID which has PFIs publicly addressable endpoint                       |
```

rest-api/README.md
```diff
@@ -0,0 +1,321 @@
...
+
```
michaelneale, line 73:
```suggestion
If the serviceEndpoint is itself a DID, this did should resolve to a document and then its serviceEndpoints can be examined for the `#pfi` item. 
```

michaelneale (changes requested):
put some suggested changes for did addressable serviceEndpoints.

rest-api/README.md
```diff
@@ -0,0 +1,321 @@
...
+| Param         | Description                   |
+| ------------- | ----------------------------- |
+| `messageType` | filters the messages returned |
+
+
+### Response
+
+| Status             | Body                       |
+| ------------------ | -------------------------- |
+| `200: OK`          | `{ data: TbdexMessage[] }` |
```
jiyoontbd, line 293: could i assume that for a completed order, the tbdexmessage array returned may look like 
`[Rfq, Quote, Order, OrderStatus, OrderStatus]`, _in that order_ ?

rest-api/README.md
```diff
@@ -0,0 +1,321 @@
...
+| `400: Bad Request` | `{ errors: Error[] }`             |
+
+### Response Body
+> [!IMPORTANT]
+> See Quote structure [here](../README.md#quote)
+
+## Submit Order
+
+### Endpoint
+`POST /threads/:thread_id/order`
```
KendallWeihe, line 222: Resurfacing a proposal from [here](https://sq-tbd.slack.com/archives/C05B6AB7JG1/p1692319359593849?thread_ts=1692307921.943449&cid=C05B6AB7JG1)

For both _order_ and _close_...

```
PUT /threads/:id

{
  "messageType": "order|close"
}
```

...and also... I would open the question of `action` rather than `messageType` (oh, or `command` per 
@phoebe-lew's comment)

----

The big thing is, no need for a `parentId` in either request. Because a `parentId` means that the client would either need to manage the state of the prior message in the thread, or else first make a call to `GET /threads/:id` as a means to retrieve the `parentId`.

The exact RESTful spec is neither here nor there, like `POST /threads/:id/order` has the same inputs as `PUT /threads/:id`+request body. But the `PUT /threads/:id` means you're exposing a way to _mutate the thread_ which is semantically different than the `POST` design which means _create an order/close_

rest-api/README.md
phoebe-lew: Probably not by default - will add a query param for `sort` so that you can get it in an ordered array

rest-api/README.md
```diff
@@ -0,0 +1,321 @@
...
+### Description
+Returns an array containing IDs of threads started by the requesting DID
+
+### Endpoint
+`GET /threads`
+
+### Response
+| Status             | Body                  |
+| ------------------ | --------------------- |
+| `200: OK.     `    | `{ data: string[]}`   |
```
jiyoontbd, line 310: thoughts on returning TbDEXMessage[][] - so we get a list of exchange, and each exchange consists of a list of tbdexmessages? 

we are thinking that having a didpay endpoint that retrieves all exchanges would be nice so we can render a list of exchanges in the 'landing page'

jiyoontbd: and also a query param so we can pass in only the exchangeIds that we need?

rest-api/README.md
```diff
@@ -0,0 +1,329 @@
...
+### Authentication
+Uses DID authn via Bearer token in header.
+
+### Endpoint
+`GET /exchanges`
+
+### Response
+| Status             | Body                  |
+| ------------------ | --------------------- |
+| `200: OK.     `    | `{ data: { string: tbdexMessage[] } }`   |
```
jiyoontbd, line 314: @phoebe-lew based on what you wrote here, if i call `GET /exchanges?ids=1,2,3`, should i expect the following?

```
{
    1 : [rfq1, quote1, order1],
    2 : [rfq2], 
    3 : [rfq3, quote3, order3, orderstatus3]
}
```

this would work. if it's easier, returning 2d array could also work
```
[
    [rfq1, quote1, order1],
    [rfq2], 
    [rfq3, quote3, order3, orderstatus3]
]
```

right now the didpay implementation assumes `tbdexMessage[][]` type as the response from this endpoint. but it can easily changed if need be to the `Map<string, tbdexMessage[]>`

rest-api/README.md
```diff
@@ -0,0 +1,329 @@
...
+| ------ | ---------------------- |
+| 400    | Validation error(s)    |
+| 400    | Failed Signature Check |
+| 404    | Thread not found       |
+| 409    | RFQ already exists     |
+| 409    | Thread already exists  |
+
+## Get Quote
+### Endpoint
+`GET /threads/:thread_id/?messageType=quote`
```
jiyoontbd, line 207: this is also descoped, because `GET /threads/:thread_id` overall is descoped below, yea?

rest-api/README.md
```diff
@@ -0,0 +1,329 @@
...
+
+---
+
+# Threads
+A thread is a series of linked tbDEX messages between Alice and a PFI for a single exchange. A thread can be created by submitting an RFQ.
+
+## Submit RFQ
+
+### Endpoint
+`POST /threads/:thread_id/rfq`
```
jiyoontbd, line 179: a couple questions -

1. a tbdex exchange always starts with an rfq. thoughts on calling this endpoint `POST /exchanges` and passing in the exchangeId (part of rfq) in the req body? this thought is similar to what @KendallWeihe proposed below, of using `PUT /threads/:thread_id` instead of `POST /threads/:thread_id/order`

2. are we aligning on using `exchanges` in place of `threads`? if so, it would be good to reflect that in submit rfq / order / close endpoint descriptions here and below.

rest-api/README.md
mistermoe: will open an issue and surface this

rest-api/README.md
mistermoe: will surface this in an issue

PR #140, opened 2023-08-18
how to do PFI discovery including ION

michaelneale:
This is a proposal to enhance how discovery is done - both in terms of "hinting" at currency pairs offered in the serviceEndpoint in the PFIs did, but also sample code of how this can be done today with the ION network anchoring. 

No specific client or server code is required, any state is stored in ION (bitcoin blockchain anchoring).

Client code can run in a browser, or server side (with or without libraries). 

cc @mistermoe related to your work https://github.com/TBD54566975/tbdex-protocol/tree/rest-api-spec

thanks to @nitro-neal for sample code for anchoring.


KendallWeihe:
Two words: Bitcoin Ordinals

But yeah, ION is good too tho

michaelneale:
@KendallWeihe yeah I started looking at that but then I worked out I would build kind of half of what we already have in ion anyway (or a buggier version thereof) and from what I read no one would like it.

discovery/README.md
```diff
@@ -0,0 +1,56 @@
...
+```bash
+npm install @decentralized-identity/ion-tools
+```
+
+
+## Sample usage
+
+```node discover.js```
+
+This will provide a list of PFIs. This uses a did:ion type of `1669` (16=P, 6=F, 9=I) to find all PFIs (and check they have a PFI type serviceEndpoint)
```
jiyoontbd, line 46: can you explain why type is 1669?

discovery/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Optional currency pair declaration
+
+To help clients find the right PFI for their needs, a PFI may also declare what currency pairs they are able to offer:
+
+```json
+      {
+        id:'pfi',
+        'type': 'PFI',
+        'serviceEndpoint': 'https://pfi.mic4.com AUD-USD,AUD-TZN,USD-BTC',
```
jiyoontbd, line 27: hmm, is this a convention we're proposing here that wallets have to be aware of, in order to crawl the DID document whose serviceEndpoint type is PFI and hint at available pairs? i thought that fetching offering  might do a better job of this.

discovery/discover.js
```diff
@@ -0,0 +1,55 @@
...
+
+  return pfiServiceEndpoints;
+}
+
+// Exporting the function to be used in other parts of your app
+export default fetchPFIServiceEndpoints;
+
+
+const endpoints = await fetchPFIServiceEndpoints();
+console.log(endpoints);
```
jiyoontbd, line 55: this bit did not work for me when i ran node discover.js - seemed to complain about the export bit and the await bit. 

changing this seemed to work though. maybe not the best way to do it but i was able to get the output i was looking for.

```javascript
exports.fetchPFIServiceEndpoints = fetchPFIServiceEndpoints;

async function main() {
  const endpoints = await fetchPFIServiceEndpoints();
  console.log(endpoints)
}

main().then(() => process.exit(0), e => { console.error(e); process.exit(1) })```

Comment (jiyoontbd) [reply discovery/discover.js:55]: oh. seems like changing the extension to `discover.mjs` worked :) same for create-and-anchor.js

File: discovery/README.md (hunk 1 above)
Comment (michaelneale) [reply discovery/README.md:46]: P is 16th letter, F is 6th, I is 9th. Just a way to remember it as the "PFI" type

File: discovery/README.md (hunk 2 above)
Comment (michaelneale) [reply discovery/README.md:27]: fetching offering does do a better job, this is a "hint" - say there are 100s returned, then it would have to fire off 100s of requests. Maybe that is ok? if there were 1000, maybe a heavy load but I guess that is a "good problem to have"? So perhaps this is over thinking it. The offers request is still the difinitive thing, so perhaps this is confusing?

File: discovery/discover.js (hunk 3 above)
Comment (michaelneale) [reply discovery/discover.js:55]: yes - needs to have a package.json with type module - I should mention that.

File: discovery/README.md (hunk 2 above)
Comment (michaelneale) [reply discovery/README.md:27]: @jiyoontbd yeah I think perhaps can remove it, keep the endpoint as a simple endpoint (be that http or whatever).

File: discovery/README.md (hunk 1 above)
Comment (jiyoontbd) [reply discovery/README.md]: sorry, i was asking why it has to be 1669 and not just `pfi` :D does it have to be a number?

**Review by jiyoontbd — APPROVED** [review]

File: discovery/README.md (hunk 1 above)
Comment (michaelneale) [reply discovery/README.md]: @jiyoontbd yes has to be a number no more than 4 digits.

File: discovery/README.md [hunk 4]
```diff
@@ -0,0 +1,90 @@
...
+
+## Registering a PFI
+
+```node create-and-anchor.mjs```
+
+A PFI will create a did (at least once) and may choose to anchor it. This is example code to show how to do that but it is not required to do it with this library. 
+
+This will create a new DID and anchor it permanently to the bitcoin blockchain via sidetree and ION. The DID (including private key) will be saved in a json file in the same directory.
+
+## Registering a PFI uses did:web
```
decentralgabe, line 82: this doesn't enable discovery, unless you use it with an ION DID, right?

discovery/create-and-anchor-did-web.mjs
```diff
@@ -0,0 +1,52 @@
...
+        type: 'EcdsaSecp256k1VerificationKey2019',
+        publicKeyJwk: authnKeys.publicJwk,
+        purposes: [ 'authentication' ]
+      }
+    ],
+    services: [
+      {
+        id:'pfi',
+        'type': 'PFI',
+        'serviceEndpoint': 'did:web:pfi.mic4.com',
```
decentralgabe, line 23: if I understand correctly this is creating an ION DID with a service endpoint pointing to a did:web DID?

I would recommend altering this to having a new verification relationship linking the two DIDs, where the did:web lists the ION DID as its controller

PR #149, opened 2023-08-30
bearer token create and verify

phoebe-lew:


js/src/crypto.ts
```diff
@@ -156,6 +179,132 @@ export class Crypto {
...
   }
+
+  /**
+   * signs the message as a compact jws with nonce
+   * @param privateKeyJwk - the key to sign with
+   * @param kid - the kid to include in the jws header. used by the verifier to select the appropriate verificationMethod
+   *              when dereferencing the signer's DID
+   * @returns the compact JWS
+   */
+  static async token(opts: TokenOptions): Promise<string> {
```
jiyoontbd, line 190: what is the difference between this method and the `sign` method in line 85?

phoebe-lew: This is for creating the bearer token for get requests and sign populates the `signature` field on a tbdex message

jiyoontbd: ah, i see they're slightly different in implementation - sign() uses `Crypto.hash(jwspayload)` while token() does not. cool!

jiyoontbd: i was originally thinking that they should be squished into one method, but that part is distinct

js/src/crypto.ts
```diff
@@ -10,14 +11,15 @@ import * as cbor from 'cbor-x'
 import { sha256 } from '@noble/hashes/sha256'
 import { Convert } from '@web5/common'
 import { EcdsaAlgorithm, EdDsaAlgorithm, Jose } from '@web5/crypto'
-import { deferenceDidUrl, isVerificationMethod } from './did-resolver.js'
+import { DidResource, deferenceDidUrl, isVerificationMethod } from './did-resolver.js'
+import { DidDocument, VerificationMethod } from '@web5/dids'
```
mistermoe, line 15:
```suggestion
```

js/src/crypto.ts
```diff
@@ -1,4 +1,5 @@
 import type { MessageModel, ResourceModel } from './types.js'
+import type { BearerTokenModel } from './pfi-rest-client.js'
```
mistermoe, line 2:
```suggestion
```

js/src/crypto.ts
phoebe-lew: eventually i wanna refactor to reduce duplication

jiyoontbd (approved):

mistermoe:
Closing this out bc we pulled this code into #147

PR #161, opened 2023-09-20
Reoganize repo

mistermoe:
thoughts? also we can probably change the name of this repo to just `tbdex`. and then each language implementation can be `tbdex-js` `tbdex-kt` `tbdex-COBOL` etc.


michaelneale:
I am ok with this - would love input from @angiejones or someone else too to double check it makes sense.

mistermoe:
@angiejones probably easiest to take a look at the [branch](https://github.com/TBD54566975/tbdex-protocol/tree/reorganize) directly.

Summary:
* updated main README to include a blurb about what the repo contains and a table of the current specs
* moved specs to `specs` directory where each spec has its own directory.
* each spec directory has a `README.md` which has all of ze content. went with 1 directory per spec to house additional images or whatever

michaelneale (approved):
seems an improvement to me, much nicer.

PR #165, opened 2023-10-06
Removing dash in orderstatus.schema.json to conform with MessageKind enum

jiyoontbd:
while working on https://github.com/TBD54566975/tbdex-kt/issues/5, we wanted to align on the naming of the order status message schema file to be consistent with the MessageKind enum, so that we can use a map to store and retrieve the JsonSchema by MessageKind enum value.


andresuribe87 (approved):
shipit

ethan-tbd (approved):

phoebe-lew (approved):

PR #177, opened 2023-10-21
replace CBOR with JCS and update spec sections

mistermoe:
## What's Changed

* Swapped CBOR for [JSON Canonicalization Scheme](https://datatracker.ietf.org/doc/html/rfc8785))
    * Using CBOR for deterministic serialization proved to be more difficult than originally anticipated specifically because of the lack of consistent implementations of the same CBOR sorting algorithm. Switched to using JSON Canonicalization Scheme defined [here in RFC 8785](https://datatracker.ietf.org/doc/html/rfc8785). 
* Updated spec sections
   *  introduced a "Common Traits" section that includes traits common across message and resource
   * updated digest section to reflect switch from CBOR -> JCS and laid out the steps needed to compute a message/resource digest
   * added section for `signatures`. List out steps needed to compute signature

### DIFF
--- a/specs/protocol/README.md
+++ b/specs/protocol/README.md
@@ -32,15 +32,8 @@ Version: Draft
     - [`private`](#private)
       - [Example Usage in RFQ message](#example-usage-in-rfq-message)
     - [`signature`](#signature-1)
-      - [Header](#header)
-        - [Supported `alg`s](#supported-algs)
-      - [Payload](#payload)
   - [ID generation](#id-generation)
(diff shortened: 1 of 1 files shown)


mistermoe:
PR that reflects changes made to spec is [available here](https://github.com/TBD54566975/tbdex-kt/pull/63)

mistermoe:
cc: @amika-sq

phoebe-lew (approved):

PR #178, opened 2023-10-24
add test vectors

mistermoe:
TODO: add mention of test vectors in root `README`

PR #187, opened 2023-11-16
Add test vectors and upload via GH action

### LINKED ISSUE #169: Include Test Vectors
Test vectors will be super helpful w.r.t. interoperability across implementations. "Test vectors" sounds fancy but it’s really as simple as tossing a `test-vectors.json` file into this repo. Json structure could be something like:

```jsonc
{
    "resources": {
        "offering": [{ /* signed offering */ }]
    },
    "messages": {
        "rfq": [{ /* signed rfq */ }],
        "quote": [{ /* signed quote */ }],
        "order": [{ /* signed order */ }],
        "orderstatus": [{ /* signed orderstatus */ }],
        "close": [{ /* signed close */ }]

    }
}
```

any/all implementations can pull this into their tests, traverse the object, and for each resource or message:
* parse
* call sign to ensure that the same signature is generated

2nd one is a bit tricky because not all signature algorithms generate deterministic signatures (`secp256k1` does). We could just include a private key in the vector file

ideally, don't want to fall into a situation where we have this combinatorial matrix of each language's sdk generating its own vectors and also consuming the vectors of every other lib

phoebe-lew:
closes #169


mistermoe:
maybe related? feel free to close or adjust as needed!

* https://github.com/TBD54566975/tbdex/issues/169
* https://github.com/TBD54566975/tbdex/pull/178

.github/workflows/publish-hosted-resources.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         uses: actions/upload-pages-artifact@v2
         with:
           # Upload entire repository
-          path: './json-schemas'
+          path: './hosted'
```
phoebe-lew, line 40: @mistermoe can you confirm that this will publish everythang? 🙏

hosted/test-vectors/message/marshal.json
```diff
@@ -0,0 +1,141 @@
...
+                            "accountHolderName": "Alfred Holder"
+                        }
+                    },
+                    "claims": [
+                        "presentation submission"
+                    ]
+                },
+                "signature": "eyJraWQiOiJkaWQ6a2V5OnpRM3NoaXpmUVlFaUI4ZHFhYzk2UDZTSFl0VUNDc1k3Z1NjaTRRZWFlYmdzZnoxVkMjelEzc2hpemZRWUVpQjhkcWFjOTZQNlNIWXRVQ0NzWTdnU2NpNFFlYWViZ3NmejFWQyIsImFsZyI6IkVTMjU2SyJ9..Y9gKs87CGalvHxlChHsL2fx2wtlPbY2-tBDciz7rSKRJkWmeIdyu-7NxwUdVMlQfJ58PTLoYPknXUKL8Hc0v4w"
+            },
+            "result": "valid"
```
KendallWeihe, line 40: Is there a paradigm over this `"result"` field? I assume the intent here is `"valid"` means that it's possible to marshal without throwing an exception, so does that imply there are other values we could put here other than `"valid"`

hosted/test-vectors/message/marshal.json
```diff
@@ -0,0 +1,141 @@
...
+                        "paymentDetails": {
+                            "phoneNumber": "+254712345678",
+                            "accountHolderName": "Alfred Holder"
+                        }
+                    },
+                    "claims": [
+                        "presentation submission"
+                    ]
+                },
+                "signature": "eyJraWQiOiJkaWQ6a2V5OnpRM3NoaXpmUVlFaUI4ZHFhYzk2UDZTSFl0VUNDc1k3Z1NjaTRRZWFlYmdzZnoxVkMjelEzc2hpemZRWUVpQjhkcWFjOTZQNlNIWXRVQ0NzWTdnU2NpNFFlYWViZ3NmejFWQyIsImFsZyI6IkVTMjU2SyJ9..Y9gKs87CGalvHxlChHsL2fx2wtlPbY2-tBDciz7rSKRJkWmeIdyu-7NxwUdVMlQfJ58PTLoYPknXUKL8Hc0v4w"
```
KendallWeihe, line 38: What's up with the two dots `..` here? Is that expected?

KendallWeihe: Oh yeah, I guess it is (just tested independently), so I assume that's a result of whatever cryptographic operations/encodings we're executing. Disregard!

KendallWeihe (approved):
lgtm ✅

hosted/test-vectors/message/marshal.json
phoebe-lew: Updated to be `errors` true/false. Used to indicate whether an error should be thrown when there is no output expected. (no output can be successful or unsuccessful, depending on the feature being tested)

hosted/test-vectors/message/marshal.json
jiyoontbd: from what i recall, this is a detached signature that excludes the body (aka where all the sensitive stuff lives in the RFQ). you can see an example of how this signature is created [here](https://github.com/TBD54566975/tbdex-js/blob/main/packages/protocol/src/crypto.ts#L124-L129)

jiyoontbd (approved):
🥳

PR #197, opened 2023-12-05
Fix RFQ example field vcs -> claims

diehuxx:
And use `private` field in RFQ example to show how what to do with PII.


michaelneale (approved):
good catch @diehuxx !

jiyoontbd (approved):
yay!

Issue #211, opened 2023-12-20
Document custom exception types 


kirahsapong:
Custom errors were introduced as part of [this tbdex-js PR](https://github.com/TBD54566975/tbdex-js/pull/113) and should be elevated here as a known set of common exceptions

PR #212, opened 2023-12-20
Add common exceptions section to README

kirahsapong:
Addresses #211

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,15 +14,13 @@
   - [tbDEX Close Message](#tbdex-close-message)
   - [tbDEX Client](#tbdex-client)
   - [tbDEX Server](#tbdex-server)
-- [Common Exceptions](#commmon-exceptions)
-  - [tbDEX Client](#tbdex-client-1)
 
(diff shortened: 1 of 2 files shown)


andresuribe87:
Relevant issue in web5-kt, which might be insisghtful, is https://github.com/TBD54566975/web5-kt/issues/144

Seeing that this is the approach in JS suggests that we should define the set of errors/exceptions in sdk-development. If that's the case, we can also ensure that we have consistent test-vectors across implementations.

phoebe-lew (approved):
love ur work

diehuxx (commented):
I think this makes more sense in `specs/http-api/README.md` since it looks like these errors only exist within the http-api, not the base protocol.

More broadly though, since errors handling is pretty differently across languages, why do we need to standardize these? It doesn't seem like the string `RequestError` is being passed in an http response. I'd prefer to let different languages handle errors in whatever way is most idiomatic.

jiyoontbd (approved):

Issue #214, opened 2023-12-21
[Proposal] Host test vectors on GH


diehuxx:
## Overview

TBDex will host test vectors on github and can be fetched and cached by implementation repos in a `.gitignore`'d place. Implementation repos will NOT need to git commit the vectors themselves; implementation repos will only commit a git ref of the `tbdex` repo containing the desired version of the test vectors.

## Hosting on Github

The github `tbdex` repo will host all test vectors for the protocol in the directory `test-vectors/protocol/` . An implementation repo can fetch the test vectors at a given git ref like so:
```sh
curl -L \                                                                                                                                                                                                                                                                                                    
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
"https://api.github.com/repos/TBD54566975/tbdex/contents/specs/?ref=$MY_GIT_REF"
```

Each vector file will follow this JSON schema:
1. `description` : A human readable description of the scenario being tested.
2. `input` : A stringified JSON TBDex object
3. `output` : If the input contains a valid TBDex object: A JSON object of the the expected parsed input.
4. `error` : boolean. True if an error is expected, false if not. No error message/detail is provided in the test vectors because different languages/implementations may handle errors quite differently. So, it's better not to be too prescriptive of implementation.

NOTE: This is nearly identical to the current test vector structure. The only difference is standardizing the type of `error`.

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "description": {
      "type": "string",
    },
    "input": {
      "type": "string",
    },
    "output": {
      "type": "object",
      "additionalProperties": true,
    },
    "error": {
      "type": "boolean"
    }
  },
  "required": ["description", "input"],
  "oneOf": [
    {
      "required": ["output"],
      "not": {"required": ["error"]}
    },
    {
      "required": ["error"],
      "not": {"required": ["output"]}
    }
  ]
}
```

### Benefits of Hosting on Github

Some reviewers will prefer the approach to hosting test vectors taken by Web5 — that is, copy-pasting each test vector file into an implementation repo. I really really don’t want to do that. Here’s my hot takes:

1. **Hosting on Github is convenient.** If an implementer wants to test against a specific version of the test vectors, they can link to a specific git ref, using [the github API](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content).
[...]

PR #215, opened 2023-12-21
#214 Restructure hosting of test vectors

diehuxx:
Separate test vectors into one test vector per file. Set up a validation script to ensure all test vectors follow a common format.

See `hosted/test-vectors/protocol/README.md` for how to use and generate test vectors

See https://github.com/TBD54566975/tbdex-js/pull/129 for how implementations can use git submodules to consume test vectors and other hosted materials.


diehuxx:
Instead of using this entire repo as a submodule in implementation repos (tbdex-js, tbdex-kt, etc.), let's pull the JSON schemas and test vectors into their own repo called tbdex-interop-suite, which is included as a submodule here and in implementation repos.

KendallWeihe:
Nice! A coupla thoughts...

- Should we also move [`scripts/test-vector-validation`](https://github.com/TBD54566975/tbdex/tree/vectors-214/scripts/test-vector-validation)?
- The [GitHub Workflow](https://github.com/TBD54566975/tbdex/blob/vectors-214/.github/workflows/publish-hosted-resources.yml) still relies on the `hosted/` directory, is this a bug?

KendallWeihe:
> Nice! A coupla thoughts...
> 
> * Should we also move [`scripts/test-vector-validation`](https://github.com/TBD54566975/tbdex/tree/vectors-214/scripts/test-vector-validation)?
> * The [GitHub Workflow](https://github.com/TBD54566975/tbdex/blob/vectors-214/.github/workflows/publish-hosted-resources.yml) still relies on the `hosted/` directory, is this a bug?

As I was typing that out, I see you committed the files back. Sorry I may be throwing this for a loop.

hosted/test-vectors/protocol/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "scripts": {
+    "validate-vectors": "node validate-vectors.js"
```
nitro-neal, line 3: nice

nitro-neal (approved):
Awesome, once these are all merged in happy to get your help to get web5-* submodule created and working in the same fashion.

Also very soon I think the unit test xml will get upload for tbdex-* and we can add them to the checkmark page in sdk-development

mistermoe (approved):
💯

PR #216, opened 2023-12-22
Update readme to detail JWT token requirement for get exchanges endpoint auth 

### LINKED ISSUE #210: [http-api] request tokens should be JWTs
currently, the request token used to authenticate requests like `GET /exchanges` is a JWS that include:

**Header**:
* `alg`
* `kid`

**Payload**:
* `timestamp`

I'm proposing that we shift to using [JWTs](https://datatracker.ietf.org/doc/html/rfc7519) instead.

 the JWT should contain the following:

**Header**:
* [`typ`](https://datatracker.ietf.org/doc/html/rfc7519#section-5.1): `JWT`.
   * we have all sorts of signatures and tokens flying around. setting `typ` to `JWT` as recommended by the JWT spec to provide a means to disambiguate
* [`kid`](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.4): fully qualified verificationMethod ID. used to locate the verification method in a DID Document that should be used to verify the JWT signature
* [`alg`](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.1): cryptographic algorithm used to compute the JWT signature

**Payload**:

* [`aud`](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3): the intended PFI's DID
  * In our existing setup, if a request token is compromised, it could potentially be misused with _any_ PFI. By incorporating the `aud` claim, we can limit the risk to just one PFI, thereby reducing the surface area for misuse.
* [`iss`](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1):  the requester's DID 
  * including to be informative. this is technically duplicative as the `kid` also includes the requester's DID 
* [`exp`](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4): expiration timestamp
  * limits the amount of time a compromised request token can be used. our current request tokens have no expiration. if one were compromised they could be used indefinitely. including `exp` limits the duration of the attack vector
* [`iat`](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.6): when the JWT was created
  * including to be informative 
* [`jti`](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.7): used as a nonce to prevent replay attacks.
   * spec text should include more detail around exactly _how_ to prevent replay attacks 

TODOs:
PR to address this issue should:
- [ ] update relevant [spec text](https://github.com/TBD54566975/tbdex/tree/main/specs/http-api#authentication-2)
- [ ] re-generate test vectors.
- [ ] publish draft PR and await PRs for [JS](https://github.com/TBD54566975/tbdex-js/issues/122) and [KT](https://github.com/TBD54566975/tbdex-kt/issues/121)

jiyoontbd:
closes #210


mistermoe:
https://github.com/TBD54566975/tbdex-js/pull/126

specs/http-api/README.md
```diff
@@ -323,18 +367,39 @@ Uses DID authn via Bearer token in header.
...
-
-### Authentication
-Uses DID authn via Bearer token in header.
+Returns an array of tbdex message arrays (a list of exchanges)
 
 ### Endpoint
 `GET /exchanges`
 
+### Authentication
+A tbdex http client must send an Http Auth Header in the form of a JWT `Bearer` token.
```
andresuribe87, line 376:
```suggestion
A tbdex http client MUST set the HTTP [Authorization Header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) in order to be authenticated. The value of the header MUST use the [Bearer Authentication Scheme] as defined in [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750), and the access token MUST be a JWT.
```

specs/http-api/README.md
```diff
@@ -323,18 +367,39 @@ Uses DID authn via Bearer token in header.
...
+
+The JWT must consist of JWS header and JWT payload containing the below fields, and signed by the client's DID.
```
andresuribe87, line 378: Aligning with the terminology as defined in https://www.rfc-editor.org/rfc/rfc7519#section-2. 

Also making it clear how the signing happens. It sounds super confusing, so may I could see it being separated into another section (or simply removed)

```suggestion
The JWT must consist of JOSE header and JWT Claims Set containing the below fields, and signed using the verification material contained in a verification method that's published as the authentication verification relationship within the client's published DID document.
```

specs/http-api/README.md
```diff
@@ -323,18 +367,39 @@ Uses DID authn via Bearer token in header.
...
+
+#### JWS Header
```
andresuribe87, line 380:
```suggestion
#### JOSE Header
```

specs/http-api/README.md
```diff
@@ -323,18 +367,39 @@ Uses DID authn via Bearer token in header.
...
+| Param  | Value | Description | 
+|--------|-------|-------------|
+| `typ`  | `JWT` | Token type  |
+| `alg`  | `HS256` | Algorithm |
+| `kid`  | `did:dht:alice...#dwn-sig` | Key ID of the sender's DID |
+
+#### JWS Payload
```
andresuribe87, line 387:
```suggestion
#### JWT Claims Set
```

specs/http-api/README.md
jiyoontbd: I'd prefer to leave this as JWS header, because from what I'm reading, JOSE can include many kinds of headers like JWS, JWE, JWK, and JWA.

But, to describe the request token header, I want to specify that it must be a `JWS` header, which must include `alg`, `typ`, and `kid` fields.

specs/http-api/README.md
```diff
@@ -323,18 +367,39 @@ Uses DID authn via Bearer token in header.
...
+| Param  | Value | Description | 
+|--------|-------|-------------|
+| `aud`  | `did:dht:pfi` | DID of the audience (i.e. PFI DID)  |
+| `iss`  | `did:dht:alice` | DID of the issuer (i.e. Alice DID) |
+| `exp`  | `1703222687` | Time when the JWT expires. 1 minute after issuance  |
+| `iat`  | `1703222795` | Time when the JWT was issued. |
+| `jti`  |  `0189f7e5-c883-7106-8272-ccb7fcba0575` | Unique identifier for the JWT. Used to prevent reply attacks. Must be uuidv7. |
```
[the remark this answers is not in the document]
jiyoontbd: > It's unclear to me how this would prevent replay attacks

@mistermoe and i chatted about this briefly. If the PFI wanted to prevent replay attacks, they'd want to keep the `jti` fields of each request in a cache / log, and check whether the same `jti` fields are sent by subsequent requests. If a token with same `jti` is sent again, then this could be an attacker using the same token to perform actions not intended by the original token sender.

To reduce the amount of `jti` the PFI would have to store at any given time, we also have `exp` field in the claimsset, which we've tentatively set to 1 min from current time. A PFI could implement a TTL cache so that any expired token falls off the cache automatically.

> Who is responsible for generating this nonce?

We would apply this spec to building tbdex http-client library, specifically, in the implementation of `TbdexHttpClient.generateRequestToken()` method, which generates a Bearer JWT token which is then used to make a call to, say, `TbdexHttpClient.getExchanges()`.

> How does the verifier (i.e. the PFI) know/trust the method used for the nonce generation?

In tbdex http-client, we were thinking of using an open source uuidv7 generation library. I was looking at [this one](https://github.com/f4b6a3/uuid-creator?tab=readme-ov-file) for tbdex http-client in kotlin.

specs/http-api/README.md
jiyoontbd: thank you! that's quite thorough, but also a bit confusing to read to me too. i'll look into how i can re-work this.

angiejones:
I take it, the [Quote guide](https://developer.tbd.website/docs/tbdex/wallet/receive-quote/#polling-for-quote) needs to be updated?

specs/http-api/README.md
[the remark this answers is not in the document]
andresuribe87: Thanks for the details!

It would appear to me that an attacker can still replay the message by waiting long enough for the cache to evict the token. Am I understanding this correctly?

Without understanding the details, I think this is being used to guarantee idempotency. A side effect of idempotency on an endpoint is that replay attacks become meaningless. Is the motivation here to address idempotency of the api, to prevent replay attacks, or to solve both?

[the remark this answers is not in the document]
jiyoontbd: > an attacker can still replay the message by waiting long enough for the cache to evict the token

they could, but during the `verifyRequestToken()` stage after the PFI receives the token for an endpoint, the cache check would not prevent the replay since it will have fallen off the cache, but an expiry check would still prevent the replay. i wrote a bit more about what verifyRequestToken does [here](https://github.com/TBD54566975/tbdex/pull/216/files#diff-1ab7678648ffe33c3535373210080ab41cd2c38573890fdfcf008700ef14c3c7R215-R218)

the motivation here is to solve for both idempotency and replay attack prevention

specs/http-api/README.md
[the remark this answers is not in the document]
jiyoontbd: referencing rfc6750 per this [other comment](https://github.com/TBD54566975/tbdex/pull/216#discussion_r1437097719)

jiyoontbd:
> I take it the [Quote guide](https://developer.tbd.website/docs/tbdex/wallet/receive-quote/#polling-for-quote) needs to be updated?

@angiejones the guide you linked should not need to be updated, since these token generation and verification is done internally by the appropriate TbdexHttpClient and TbdexHttpServer methods. But it might be a good thing for a PFI developer needs to be aware of (especially how to verify a bearer jwt token), in case they need to build a tbdex server without the provided tbdex client/server libraries.

mistermoe (approved):
awesome work @jiyoontbd!

angiejones:
> > I take it the [Quote guide](https://developer.tbd.website/docs/tbdex/wallet/receive-quote/#polling-for-quote) needs to be updated?
> 
> @angiejones the guide you linked should not need to be updated, since these token generation and verification is done internally by the appropriate TbdexHttpClient and TbdexHttpServer methods. But it might be a good thing for a PFI developer needs to be aware of (especially how to verify a bearer jwt token), in case they need to build a tbdex server without the provided tbdex client/server libraries.

@chris-tbd please review the changes here and decide if you want to add any text to the the PFI guides

diehuxx (approved):
noice!

PR #221, opened 2024-01-16
Account Balance spec text

### LINKED ISSUE #209: [http-api] Consider introducing an endpoint that can be used to get a balance
Formally, the concept of an "account" does not exist at the tbdex layer nor is it necessary IMO. The concept of an account however may be beneficial with respect to PFIs. An account needn't be anything complicated. They can be represented as individual DIDs. Technically this already exists because `GET /exchanges` is effectively returning all exchanges for a given "account" where that account is represented as a DID. 

an endpoint like `GET /balances` could be used to fetch balances that Alice has with a PFI. 

For example, If a PFI had a  `USD -> USDC_STORED_BALANCE` offering, Alice would be purchasing USDC _to be custodied_ with the PFI. Alice could then use `GET /balances` to get her balance. 

The PFI could then have another offering like `USDC_STORED_BALANCE -> MXN` which could be used to off-ramp her stored balance. 

This also provides a standarized means to facilitate institutional top-ups

> [!NOTE]
> incorporating something like this would by no means force all PFIs to act as custodians. Rather, it provides the ability for those who choose to provide that functionality.

phoebe-lew:
closes #209 

still need: 

- [x] type for balance at protocol level
- [x] whether it's a resource or not


specs/http-api/README.md
```diff
@@ -345,7 +350,43 @@ Uses DID authn via Bearer token in header.
...
+---
+
+## List Balances
+This endpoint is *OPTIONAL*. It is only relevant for PFIs which expose a stored balance functionality either for institutional or retail customers.
+
+### Description
+Returns an array of balance amounts for each currency held by the caller.  
+
+### Authentication
+Uses DID authn via Bearer token in header.
```
diehuxx, line 362: We've got this `protected` language written for other endpoints as a shorthand for bearer tokens
```suggestion
### Protected
True
```

mistermoe:
@diehuxx thoughts on whether balances resp should be a "resource"

specs/protocol/README.md
mistermoe, line 1: do we want to call out that _any_ PFI that offers `STORED_BALANCE` as a `payout` kind must _necessarily_ have a respective offering with `STORED_BALANCE` as a `payin` kind?

hosted/json-schemas/balance.schema.json
```diff
@@ -0,0 +1,34 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://tbdex.dev/offering.schema.json",
+    "definitions": {
+      "Balance": {
+        "type": "object",
+        "additionalProperties": false,
+        "properties": {
```
diehuxx, line 8: Maybe also an optional "name" or "description" field. We can also discuss further in another PR if you'd prefer to close this out ASAP.

diehuxx (approved):

specs/http-api/README.md
```diff
@@ -476,5 +481,41 @@ True
...
+### Endpoint
+`GET /balances`
+
+### Response
+| Status             | Body                                   |
+| ------------------ | -------------------------------------- |
+| `200: OK.     `    | `{ data: { Balance[] } }` |
+| `400: Bad Request` | `{ errors: Error[] }`                  |
+| `404: Not Found`   | N/A                                    |
+| `403: Forbidden`   | N/A                                    |
```
diehuxx, line 502: why 403 instead of 401?

diehuxx: ah wait i just read up more on the difference and i agree now. 401 == "idk u", 403 == "get outta here jerry"

specs/protocol/README.md
```diff
@@ -222,6 +232,42 @@ An `Offering` is used by the PFI to describe a currency pair they have to _offer
...
+    "from": "did:ex:pfi",
+    "kind": "balance",
+    "id": "balance_01ha82y8d0fhstg95hhfjwmgxf",
+    "createdAt": "2023-09-13T20:15:22.528Z",
+    "updatedAt": "2023-09-13T20:15:22.528Z"
+  },
+  "data": {
+    "balances": [
+      {
+        /** ISO 4217 currency code or widely adopted cryptocurrency code as string */
```
michaelneale, line 256: drives me wild that json can't have comments 😂

michaelneale (approved):
I like the look of this. Only things that gave be pause was that this is the 2nd resource - and its a very (heavily) authed one (not really a problem, just an observation). The other is the reserved payment type "stringly typed" thing, don't have any other thoughts on that but it was only thing I paused over for a second. Nice.

specs/http-api/README.md
michaelneale: yeah 403 is "don't even try - you don't have access". 401 usually means you need to identify yourself and you may get access. Sometimes 404 is also used as you don't want to give away negative existence of a resource (but in this case it is supported or not, so not giving anything away).

hosted/json-schemas/balance.schema.json
phoebe-lew: Do you think there's more description that would be useful beyond the currency and amount?

phoebe-lew:
> Only things that gave be pause was that this is the 2nd resource - and its a very (heavily) authed one

@michaelneale agreed, we went back and forth on whether it should be a resource quite a bit. I think maybe a third category of "thing" will emerge eventually (besides `resource` and `message`), but at this point we thought it was easier to just roll it into resource instead of trying to come up with a new concept.

specs/protocol/README.md
```diff
@@ -616,6 +662,62 @@ Quick explanation of terms used.
 | payout currency | currency that the PFI is paying out to Alice. Alice will _receive_ the payout currency from the PFI.                                                               |
 | payin currency  | currency the PFI will accept in exchange for the payin currency. The PFI will _receive_ the payin currency from Alice.                                             |
 
+# Stored Balances
+Stored balances are assets or funds custodied by a PFI on behalf of a customer. They are relevant for PFIs which provide options for institutional top ups, or which are custodial. Supporting stored balances is entirely optional. The intent is to provide the ability for PFIs who choose to do so.
+
+If a PFI offers `STORED_BALANCE` as a payout kind, they MUST necessarily have a respective offering with`STORED_BALANCE` as a payin kind. 
+
+## Top up stored balance
```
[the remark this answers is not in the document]
diehuxx: The former. A balance can be topped up, which will update the `updatedAt`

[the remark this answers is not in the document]
phoebe-lew: @kirahsapong go for it!

specs/http-api/README.md
```diff
@@ -476,5 +481,41 @@ True
...
+
+#### Example
+
+```json
+{
+  "data": [
```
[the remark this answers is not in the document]
phoebe-lew: Hey @kirahsapong , it should return an array of Balance objects which each look like this, similar to how an exchange is a list of tbdex messages, each of which are signed independently:

```json
{
  "metadata": {
    "from": "did:ex:pfi",
    "kind": "balance",
    "id": "balance_01ha82y8d0fhstg95hhfjwmgxf",
    "createdAt": "2023-09-13T20:15:22.528Z",
    "updatedAt": "2023-09-13T20:15:22.528Z"
  },
  "data": {
    "balances": [
      {
        /** ISO 4217 currency code or widely adopted cryptocurrency code as string */
        "currencyCode": "USD", 
        /** same format used to represent currency values across messages */
        "available": "400.00",
      },
      {
        "currencyCode": "MXN", 
        "available": "100.00",
      }
    ]
  },
  "signature": "eyJhbGciOiJFZERTQSIsImtpZCI6ImRpZDprZXk6ejZNa2syc1QyZUtvQWdUUTdzWjY3YTdmRDMzR21jYzZ1UXdaYmlxeWF5Rk1hYkhHI3o2TWtrMnNUMmVLb0FnVFE3c1o2N2E3ZkQzM0dtY2M2dVF3WmJpcXlheUZNYWJIRyJ9..9EBTL3VcajsQzSNOm8GElhcwvYcFGaRp24FTwmC845RCF84Md-ZB-CxdCo7kEjzsAY8OaB55XFSH_8K9vedhAw"
}
```

I'll remove the example here and link to the one in the `protocol` spec to consolidate source of truth.

PR #222, opened 2024-01-16
Callbacks spec update

### LINKED ISSUE #167: Introduce Webhooks / Callbacks
# Motivation
Currently, in order to receive new messages for a given exchange, Alice must poll the respective PFI's http api. A handful of people have brought up adding webhook / callback support.

Two approaches came to mind that may work.

## 1. `replyTo` property
We could consider adding a `replyTo` property to `rfq.metadata` or more generally `message.metadata`. `replyTo` is a fully qualified URI which effectively acts as a callback URL / webhook. The value of `replyTo` could either be a DID or just a URL.

if `replyTo` is present, a PFI would send any/all new messages for a given exchange to the supplied URL

## 2. DID service endpoint
Alternatively, this could happen implicitly by resolving the sender's DID and looking for a predefined service endpoint type. if the predefined service endpoint is present, a PFI would send any/all new messages for a given exchange to that service endpoint.

---

In order to support either approach, we'll need to specify how the webhook url works. e.g.
* which http method?
* how does authn work?
    *  could happen implicitly since every tbdex message is signed and alice presumably has record of the exchange any incoming message is in response to

phoebe-lew:
closes #167


specs/http-api/README.md
```diff
@@ -196,34 +208,65 @@ No authorization required. Offerings are publicly accessible
...
-No Authorization required to submit an RFQ
+No Authorization required to create an exchange
 
 ### Request Body
+
+#### `CreateExchangeRequest`
+| field            | data type | required | description                                                                                       |
+| ---------------- | --------- | -------- | ------------------------------------------------------------------------------------------------- |
+| `rfq`           | object    | Y        | The request for quote                       |
+| `replyTo` | object    | N        | A string containing a valid URI where new messages from the PFI will be sent |
```
jiyoontbd, line 227:
```suggestion
| `replyTo` | string    | N        | A string containing a valid URI where new messages from the PFI will be sent |
```

specs/http-api/README.md
```diff
@@ -156,6 +157,17 @@ The IDs of individual tbDEX messages are used as idempotency keys
...
 
+# Callbacks
+Callbacks are implemented via the `replyTo` property on the [request to create an exchange](#create-exchange). 
+
+`replyTo` is a fully qualified URI which could either be a DID or just a URL.
+
+If `replyTo` is present, a PFI will send any/all new messages for a given exchange to the supplied URI. This makes the URI scoped to each exchange, allowing the caller to specify a different URI per exchange if they so wish. 
+
+If `replyTo` is _not_ present, the caller will have to poll the PFI for the exchange in question to receive new messages.
+
```
jiyoontbd, line 168: just a general q about replyTo - does the URI change depending on whether Alice is talking tbdex via a self-custodial vs custodial wallet?

specs/http-api/README.md
```diff
@@ -196,34 +208,65 @@ No authorization required. Offerings are publicly accessible
...
-| 400    | Validation error(s)     |
-| 400    | Failed Signature Check  |
-| 404    | Exchange not found      |
-| 409    | RFQ already exists      |
-| 409    | Exchange already exists |
+| Status | Description                  |
+| ------ | -----------------------------|
+| 400    | Validation error(s)          |
+| 400    | Failed Signature Check       |
+| 409    | Exchange already exists      |
```
jiyoontbd, line 269: do we want to still keep the 404 exchange not found?

diehuxx: No, since this endpoint is what creates an exchange, there should not be an existing exchange with the same `exchange_id`

specs/http-api/README.md
```diff
@@ -196,34 +208,65 @@ No authorization required. Offerings are publicly accessible
 # Exchanges
 An exchange is a series of linked tbDEX messages between Alice and a PFI for a single exchange. An exchange can be created by submitting an RFQ.
 
-## Submit RFQ
+## Create Exchange
 
 ### Endpoint
-`POST /exchanges/:exchange_id/rfq`
+`POST /exchanges/:exchange_id`
```
diehuxx, line 214: We could take it a step further. `exchange_id` is already in the RFQ. It's superfluous in the URL.

```suggestion
`POST /exchanges`
```

jiyoontbd: +1 to this - could this actually be applied to all our POST /exchanges/xyz endpoints, since all tbdex message metadata contain exchangeId?

phoebe-lew: I'm a fan of removing it, @mistermoe any opposition?

specs/http-api/README.md
phoebe-lew: The URI can actually be different per exchange. In general, no difference in terms of how the callback is used if coming from a self-custodied vs custodial wallet though.

specs/http-api/README.md
phoebe-lew: Going to open a separate issue for this because it'll require refactoring httpclient/httpserver in JS/KT outside the context of implementing callbacks

diehuxx (approved):

PR #224, opened 2024-01-17
Add `replyTo` to RFQ schema

jiyoontbd:


jiyoontbd:
this is incorrect - we are not making changes to the tbdex message at the protocol layer, but at the http request body layer.

PR #227, opened 2024-01-19
Clarify Offering requiredPaymentDetails undefined

diehuxx:


specs/protocol/README.md
```diff
@@ -115,7 +115,7 @@ An `Offering` is used by the PFI to describe a currency pair they have to _offer
 | field                    | data type                               | required | description                                                                                       |
 |--------------------------|-----------------------------------------|----------|---------------------------------------------------------------------------------------------------|
 | `kind`                   | string                                  | Y        | Type of payment method (i.e. `DEBIT_CARD`, `BITCOIN_ADDRESS`, `SQUARE_PAY`)                       |
-| `requiredPaymentDetails` | [JSON Schema](https://json-schema.org/) | N        | A JSON Schema containing the fields that need to be collected in order to use this payment method |
+| `requiredPaymentDetails` | [JSON Schema](https://json-schema.org/) | N        | A JSON Schema containing the fields that need to be collected in the RFQ's selected payment methods in order to use this payment method. If `requiredPaymentDetails` is blank, then the RFQ's selected payment method will always be accepted. |
```
jiyoontbd, line 118: this feels a bit odd to me. does this mean that Alice can send in _anything_ in her payin payment method and the PFI will accept it? like, Alice can buy bitcoin with cheese?

diehuxx: > like, Alice can buy bitcoin with cheese?

Youre describing my hopes and dreams.

> does this mean that Alice can send in anything in her payin payment method

In my opinion, yes Alice can send anything in the payment details. If the Offering does not have any required payment details, there are no requirements on the payment details contained in Alice's RFQ. If the PFI does not want to accept cheese as payment, they better ~go to another farmers market~ specify that in their Offering.

In practice, Alice would realistically just leave the payment details empty if there were no required payment details.

Is there a different behavior that you would prefer for absent `requiredPaymentDetails`?

jiyoontbd: blue cheese for bitcoin is a fair trade tbh. they're both funky and not everyone likes them!

anyways, back to the main point - mostly i am not sure why requiredPaymentDetails is allowed to be optional. Shouldn't the PFI always have some preferred method of payin payment they accept for what they're paying out?  

another question would be, does the lack of this field also imply that Alice can get bitcoin without paying the PFI anything?

diehuxx: Yes, it implies that Alice does not have the pay the PFI if the PFI does not require it. It is up to the PFI.

jiyoontbd (approved):

PR #230, opened 2024-01-27
Render message sequence in Mermaid

quad:
If the sequence diagram ever needs to change, perhaps Mermaid will be easier to iterate upon than a handmade Excalidraw.


jiyoontbd (approved):
thank you!

diehuxx (approved):

PR #232, opened 2024-01-29
Strongly type sensitive data in RFQ

diehuxx:
Previously, we considered the concept of a top-level `private` property in RFQ, which would contain the private data specified by the Offering, where corresponding `Rfq.data` would be hashed. This has proved to be complicated in design and implementation.

### Proposal
The potentially sensitive fields are `claims`, `payinMethod.paymentDetails`, and `payoutMethod.paymentDetails`. These fields will NEVER be part of the signed payload. In other words, when creating a signature for an RFQ, the sensitive fields are stripped from the payload and then signed.

We also add new required fields in the `Rfq.data`: `claimsHash`, `payinMethod.paymentDetailsHash`, and `payoutMethod.paymentDetailsHash`. As the name implies, these fields are a hash of their counterpart. The hashed fields ARE part of the signed payload, the same as other fields like `offeringId`, `payinAmount`, etc.

### TODO
- [ ] Update Test Vectors
- [ ] Javascript implementation
- [ ] Kotlin Implementation

### DIFF
--- a/specs/protocol/README.md
+++ b/specs/protocol/README.md
@@ -258,49 +258,6 @@ The `metadata` object contains fields _about_ the message and is present in _eve
 ### `data`
 The actual message content. This will _always_ be a JSON object. The [Message Kinds section](#message-kinds) specifies the content for each individual message type
 
-
-### `private`
-Often times, an RFQ will contain PII or PCI data either within the `credentials` being presented or within `paymentDetails` of `payinMethod` or `payoutMethod` (e.g. card details, phone numbers, full names etc). 
-
-In order to prevent storing this sensitive data with the message itself, the value of a property containing sensitive data can be a [hash](#Hashing) of the sensitive data. The actual sensitive data itself is included in the `private` field. 
-
-The `private` field is ephemeral and **MUST** only be present when the message is initially sent to the intended recipient
-
-The value of `private` **MUST** be a JSON object that matches the structure of `data`. The properties present within `private` **MUST** only be the properties of `data` that include the hash counterpart.
-
-> **Note**
-> Rationale behind the `private` JSON object matching the structure of `data` is to simplify programmatic hash evaluation using JSONPath to pluck the respective hash from `data`. **NOTE**: we should try this to make sure it's actually "easy"
-
-#### Example Usage in RFQ message
-
-```json
-{
-  "data": {
-    "offeringId": <OFFERING_ID>,
-    "payoutAmount": "STR_VALUE",
-    "claims": <PRESENTATION_SUBMISSION_HASH>, <---- hash
-    "payinMethod": {
-      "kind": "BTC_ADDRESS",
-      "paymentDetails": <OBJ_HASH> <---- hash
-    },
-    "payoutMethod": {
-      "kind": "MOMO_MPESA",
-      "paymentDetails": <OBJ_HASH> <---- hash
-    }
-  },
-  "private": {
-    "claims": <PRESENTATION_SUBMISSION>, <---- actual
-    "payinMethod": {
-      "paymentDetails": <OBJ> <---- actual
-    },
-    "payoutMethod": {
-      "paymentDetails": <OBJ> <---- actual
-    }
-  }
-}
-```
-
-
 ### `signature`
 see [here](#signatures) for more details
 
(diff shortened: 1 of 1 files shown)


specs/protocol/README.md
```diff
@@ -319,19 +276,29 @@ Currency amounts have type `DecimalString`, which is string containing a decimal
...
+|----------------|---------------------------------------------------|----------|-----------------------------------------------------------------------------------------------------------------------------|
+| `offeringId`   | string                                            | Y        | Offering for which Alice would like to get a quote                                                                          |
+| `payinAmount`  | [`DecimalString`](#decimalstring)                 | Y        | Amount of payin currency you want in exchange for payout currency                                                           |
+| `claims`       | string[]                                          | N        | An array of claims that fulfill the requirements declared in an [Offering](#offering). Excluded from the signature payload. |
+| `claimsHash`   | string                                            | Y        | A hash [digest](#digest) of `claims`                                                                                        |
```
amika-sq, line 284: Should `claimsHash` be not required?

Little bit confused what `claimsHash` would be if `claims` isn't present.

specs/protocol/README.md
```diff
@@ -319,19 +276,29 @@ Currency amounts have type `DecimalString`, which is string containing a decimal
...
-| `paymentDetails` | object    | N        | An object containing the properties defined in an Offering's `requiredPaymentDetails` json schema |
+| field                | data type | required | description                                                                                                                             |
+|----------------------|-----------|----------|-----------------------------------------------------------------------------------------------------------------------------------------|
+| `kind`               | string    | Y        | Type of payment method (i.e. `DEBIT_CARD`, `BITCOIN_ADDRESS`, `SQUARE_PAY`)                                                             |
+| `paymentDetails`     | object    | N        | An object containing the properties defined in an Offering's `requiredPaymentDetails` json schema. Excluded from the signature payload. |
+| `paymentDetailsHash` | string    | Y        | A hash [digest](#digest)  of `paymentDetails`                                                                                           |
```
amika-sq, line 300: Same here. Not sure what a hash of a non-present `paymentDetails` would entail.

specs/protocol/README.md
diehuxx: Good catch -- typo on my part.

specs/protocol/README.md
diehuxx: Good catch -- typo on my part.

KendallWeihe:
> This has proved to be complicated in design and implementation.

Could you help me understand how the proposal here is a simplification? I'm guessing the complication arises during implementation with these private fields being mixed types: objects when they're _actual_ and strings when they're _hashed_, is that right?

Is there a secondary complication of mapping `private` fields onto their `data` counterparts (as in, two different tree structures where nodes are references of one another)?

From what's currently in `main`:

> The private field is ephemeral and MUST only be present when the message is initially sent to the intended recipient

I think probably we still want to include language like this, but perhaps I need to understand more of the complication first.

specs/protocol/README.md
```diff
@@ -319,19 +276,29 @@ Currency amounts have type `DecimalString`, which is string containing a decimal
...
+| `payinAmount`  | [`DecimalString`](#decimalstring)                 | Y        | Amount of payin currency you want in exchange for payout currency                                                           |
+| `claims`       | string[]                                          | N        | An array of claims that fulfill the requirements declared in an [Offering](#offering). Excluded from the signature payload. |
+| `claimsHash`   | string                                            | N        | A hash [digest](#digest) of `claims`                                                                                        |
+| `payinMethod`  | [`SelectedPaymentMethod`](#selectedpaymentmethod) | Y        | Specify which payment method to send payin currency.                                                                        |
+| `payoutMethod` | [`SelectedPaymentMethod`](#selectedpaymentmethod) | Y        | Specify which payment method to receive payout currency.                                                                    |
+
+#### Unsigned Fields
```
decentralgabe, line 288: This name is possibly misleading. I would instead refer to it as `Disclosable Fields` as you are proposing to sign over hashes and then disclose the unsigned preimage.

decentralgabe:
two thoughts...

a meta thought is that we should have an issue to do discussion / find consensus before a PR 

the second thought is that if we switched to using [SD-JWT](https://datatracker.ietf.org/doc/draft-ietf-oauth-selective-disclosure-jwt/) we would get this functionality for free 

in absence of using SD-JWT we're forced to kind of half-re-specify SD-JWT. if we're going down that route we should look into their security considerations, such as using salted hashes, and adding noise (dummy hashes).

selfissued:
@decentralgabe anticipated both of my comments.  I was going to say that if you're going to send hashes, they need to be salted hashes or there will be a one-to-one mapping between unhashed values and hashed values - enabling probes for known values.

I was going to say that if you want to understand the use of salted hashes in JSON data structures, have a look at [SD-JWT](https://datatracker.ietf.org/doc/draft-ietf-oauth-selective-disclosure-jwt/).  Of course, per Gabe's suggestion, if it appears to be applicable to your use case, I would simply use it, rather than reinvent parts of it.  That way, the IETF security review process is on your side.

decentralgabe:
Another consideration that needs to be added is how the undisclosed properties become disclosed. Is that a new message request/response flow?

diehuxx:
Closing this PR while I work on a new PR that implements the design laid out in [this comment](https://github.com/TBD54566975/tbdex/issues/274).

PR #241, opened 2024-02-20
Remove outdated test vector validations

diehuxx:
Removing a directory containing scripts for the old test vector validation that we no longer use. We now use scripts in the [hosted/test-vectors directory](https://github.com/TBD54566975/tbdex/tree/main/hosted/test-vectors/protocol).


KendallWeihe (approved):
clean up! clean up!

Issue #225, opened 2024-01-18
Consider removing `exchange_id` from submit message endpoints


phoebe-lew:
Per https://github.com/TBD54566975/tbdex/pull/222#discussion_r1454286342

If we decide to, create corresponding issues in language implementation repos

KendallWeihe:
I see a trend here towards two endpoints, one for resources and another for exchanges. 

1. `GET /resources`
2. `POST|PUT|GET /exchanges`

We have a requirement of making tbDEX protocol-agnostic (not married to HTTP), so anything we add to the HTTP layer is purely stylistic (AKA duplicative), with the intent of making it easier to comprehend. 

I think we should commit to one design choice or another b/c a middle ground is probably the most confusing. Either we implement the above two endpoints, without any additional HTTP parts (no query params, no path params, no headers, the bodies are 1:1 matches with message/resource types), or we duplicate information across the message/resource types and the HTTP parts with the intent of matching RESTful design patterns.

---

To contrast the above two endpoints with something more decorative:

- `GET /resources/offers`
- `POST /exchanges/rfq`
- `GET /exchanges/:id/quote`
- `POST /exchanges/:id/order`
- `GET /exchanged/:id/order-status`

diehuxx:
> We have a requirement of making tbDEX protocol-agnostic (not married to HTTP)

Agreed

> so anything we add to the HTTP layer is purely stylistic (AKA duplicative), with the intent of making it easier to comprehend.

Not sure if I agree. Though the base protocol layer should be agnostic to HTTP, I understand the HTTP layer to be nearly as important as the base protocol layer because we are expecting the HTTP protocol to be a widely used way in production to send/receive TBDex protocol objects. It's possible I'm quibbling about semantics.

> I see a trend here towards two endpoints, one for resources and another for exchanges.
> without any additional HTTP parts (no query params, no path params, no headers, the bodies are 1:1 matches with message/resource types)

This approach is pretty restrictive. Without headers for authz on appropriate endpoints, authz becomes complicated. Without query params for `GET` endpoints, it becomes difficult to do simple filtering for past exchanges or resources. The restriction that bodies are 1:1 matches of message/resource types would preclude behavior like the `replyTo` field in the current submit RFQ request body.

> or we duplicate information across the message/resource types and the HTTP parts with the intent of matching RESTful design patterns.

[...]

PR #243, opened 2024-02-21
#225 Remove exchange id from http submit urls

diehuxx:
Addresses #225

There is no reason to put the exchange id in the URL because the submitted message will always contain the exchange id

### DIFF
--- a/specs/http-api/README.md
+++ b/specs/http-api/README.md
@@ -281,7 +281,7 @@ An exchange is a series of linked tbDEX messages between Alice and a PFI for a s
 Submits an RFQ (Request For Quote). Alice is asking the PFI to provide a Quote so she can evaluate it.
 
 ### Endpoint
-`POST /exchanges/:exchange_id`
+`POST /exchanges`
 
 ### Authentication
 Refer to [Signature Verification Section]() of the tbDEX spec  
@@ -371,7 +371,7 @@ True
 Submits the Order. Alice wants to accept the Quote and execute the transaction.
 
 ### Endpoint
-`POST /exchanges/:exchange_id/order`
+`POST /exchanges/order`
 
 ### Protected
 False
@@ -401,7 +401,7 @@ False
 Closes the exchange. Indicates that Alice is no longer interested
 
 ### Endpoint
-`POST /exchanges/:exchange_id/close`
+`POST /exchanges/close`
 
 ### Protected
 False


specs/http-api/README.md
```diff
@@ -281,7 +281,7 @@ An exchange is a series of linked tbDEX messages between Alice and a PFI for a s
 Submits an RFQ (Request For Quote). Alice is asking the PFI to provide a Quote so she can evaluate it.
 
 ### Endpoint
-`POST /exchanges/:exchange_id`
+`POST /exchanges`
```
mistermoe, line 284: is this not `POST /exchanges/rfq` because an RFQ creates an exchange?

specs/http-api/README.md
```diff
@@ -371,7 +371,7 @@ True
 Submits the Order. Alice wants to accept the Quote and execute the transaction.
 
 ### Endpoint
-`POST /exchanges/:exchange_id/order`
+`POST /exchanges/order`
```
mistermoe, line 374: given that `POST /exchanges` creates a _new_ exchange using an RFQ _without_ the path pattern followed for subsequent endpoints like this one, would it make sense to use `PUT` instead of `POST`?  something feels slightly awkward about this but maybe i'm the only one feeling that way in which case ignore.

specs/http-api/README.md
diehuxx: It was `POST /exchanges/rfq` originally but we dropped the `/rfq` in [this spec change](https://github.com/TBD54566975/tbdex/pull/222) where we modified the request body of that endpoint to take a callback url in the `replyTo` field.

I'm not strongly opinionated about keeping or dropping `/rfq`, but I don't want to flip-flop on it more than twice.

specs/http-api/README.md
diehuxx: Agreed, `POST` seems correct for both `POST /exchanges` and `POST /exchanges/order`

diehuxx: > something feels slightly awkward about the proposed path but maybe i'm the only one feeling that way in which case ignore.

I'm def open to other ideas, or if this proposal just doesn't feel right, it's no biggie if we reject this one.

KendallWeihe: I added some thoughts here https://github.com/TBD54566975/tbdex/issues/225

specs/http-api/README.md
jiyoontbd: we're making an assumption that RFQ is always the first message to kick off an exchange so that's why we dropped the `/rfq` bit in the previous spec change - however my eye does twitch a little seeing that this is the only endpoint that does not specify the message kind like `POST /exchanges/...` in the uri.

phoebe-lew (approved):
LGTM, also discussed briefly with @mistermoe and @jiyoontbd that removing `/rfq` makes sense because of the inclusion of the `replyTo` field.

`POST /exchange/kind` for remaining message types also makes sense

diehuxx:
Closing since we landed on a different proposal
https://github.com/TBD54566975/tbdex/issues/225#issuecomment-1986057713

PR #244, opened 2024-02-22
Add `protocol` field to protocol spec

### LINKED ISSUE #159: add `version` property to `metadata` for both `Resource` and `Message`
related: #155

kirahsapong:
closes #159 

Implements proposal [here](https://hackmd.io/TucOUP8UQt-XKj8G9UsJ2Q?both)


specs/protocol/README.md
```diff
@@ -81,6 +81,7 @@ The `metadata` object contains fields _about_ the resource and is present in _ev
 | `id`        | Y              | The resource's ID                           |
 | `createdAt` | Y              | ISO 8601 timestamp                          |
 | `updatedAt` | N              | ISO 8601 timestamp                          |
+| `protocolVersion` | Y        | Version of the protocol in use (x.x format) |
```
mistermoe, line 84: do we foresee including other versions? if not, thoughts on using just `version`?

KendallWeihe: Hmmmm naming is hard. Valid point's @kirahsapong! 

However, I think it would be unwise to venture into the realm of versioning messages & resources independently such that we have to consider compatibility across message versions & resource versions (imagine a world where an exchange is transpiring where an RFQ is of one version but then the Quote is of another version and somehow the software has to reconcile that.

So, I think it's safe to assume we'll version the protocol as a whole, which consists of all the message types and resources for that given version. I vote `version` is sufficient in conveying the intent here.

KendallWeihe: oh yeah good point, interpretability ... meh makes sense to me but I'm not an outsider looking in so I have a bias

diehuxx: > there prob still remains a world in which version support drops for one party halfway through an exchange,

This should not happen. We should note in the spec that version must remain the same across messages in a given exchange. I.e. if the `exchangeId` is the same, then `version` must be the same.

diehuxx: @kirahsapong Not sure I understand. What's orthogonal to version support dropping halfway through an exchange? I'm attempting to define how to handle version support changing halfway through an exchange.

diehuxx: I think i'm confused about what version support dropping halfway through an exchange actually means. Can you give an example of a scenario where this happens?


I imagine two parties would begin an exchange with a given version only if they support exchanges at that version of TBDex.

I can see a situation where Alice submits an RFQ to a PFI with a version that the PFI does not support. In that case, the PFI should return a 4XX response, not a Close message of a version that they don't support.

phoebe-lew: +1 on `protocol`. 

For dropping support for a version, I suppose that's within the power of a PFI or a wallet to do mid-exchange. (And if we assume that there are a lot of exchanges happening simultaneously, it is almost unavoidable that a PFI/wallet may have exchanges for two different versions unless they do a hard cut over on version.)

We should think through this a bit more and go through the scenario of migrating from one version to another, to come up with a recommended strategy that can be added to the spec text.

hosted/test-vectors/protocol/vectors/parse-close.json
```diff
@@ -1,19 +1,20 @@
 {
   "description": "Close parses from string",
-  "input": "{\"metadata\":{\"from\":\"did:dht:abecscpi377h647nj5j8oz6yq544i8bzdze83mxsghskahq35ddy\",\"to\":\"did:ex:alice\",\"exchangeId\":\"abcd123\",\"kind\":\"close\",\"id\":\"close_01hpsc6aqten79yhw1gg25prmj\",\"createdAt\":\"2024-02-16T16:32:31.482Z\"},\"data\":{\"reason\":\"The reason for closing the exchange\"},\"signature\":\"eyJhbGciOiJFZERTQSIsImtpZCI6ImRpZDpkaHQ6YWJlY3NjcGkzNzdoNjQ3bmo1ajhvejZ5cTU0NGk4YnpkemU4M214c2doc2thaHEzNWRkeSMwIn0..LfYrchKuy0KtdRl-HRgLDSwRvzVAz6mApuQdJU5prOchLMRJkzcjROWxn5JbzkwVajH6VN4Xmhv6UcOfmZmoCA\"}",
+  "input": "{\"metadata\":{\"from\":\"did:dht:6jcq5i9guc85aucgyymfo7wbbqfbb16yy13m8xoqguahy6u6ndiy\",\"to\":\"did:ex:alice\",\"exchangeId\":\"rfq_01hq7g4hrrev4v12t7r3t53ayz\",\"kind\":\"close\",\"id\":\"close_01hq7g4hrrev4v12t7r7pnwgsm\",\"createdAt\":\"2024-02-22T04:10:49.496Z\",\"protocolVersion\":\"0.26\"},\"data\":{\"reason\":\"The reason for closing the exchange\"},\"signature\":\"eyJhbGciOiJFZERTQSIsImtpZCI6ImRpZDpkaHQ6NmpjcTVpOWd1Yzg1YXVjZ3l5bWZvN3diYnFmYmIxNnl5MTNtOHhvcWd1YWh5NnU2bmRpeSMwIn0..f_Y-FEcU8KyBDbWJNC4fHyRTZWksQtDYcyE49EI7OrUmBxCxO2W9Hc6qZUhlkUonV7p3OibWYM_evUlIXsxWCg\"}",
```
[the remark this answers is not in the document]
jiyoontbd: @kirahsapong i am going to bet good money that kt tests fail. i would win this bet coz i ran the tests 

an issue to tackle later has been filed in tbdex-kt https://github.com/TBD54566975/tbdex-kt/issues/180

my changes that fixes issues in kt tests is here - would like to request this be merged first 🙏 https://github.com/TBD54566975/tbdex/pull/247

diehuxx (approved):
This looks great! I know merging is pending implementation, so tag me when those are ready and I'll get those reviewed as soon as possible.

KendallWeihe:
I think we need to keep the test vectors using `did:dht`'s rather than `did:jwk`'s https://github.com/TBD54566975/tbdex/pull/247#issuecomment-1962188385

phoebe-lew (approved):

PR #246, opened 2024-02-22
Update test vec generation recommendation

diehuxx:
My opinion

### DIFF
--- a/hosted/test-vectors/protocol/README.md
+++ b/hosted/test-vectors/protocol/README.md
@@ -17,5 +17,5 @@ Each vector file must follow the schema in `vectors.json.schema`.
 3. Validate the test vector structure.
   a. Set up with `cd hosted/test-vectors/protocol & npm install`.
   b. Run with `npm run validate-vectors`.
-4. Test the updated vectors against an implmentation of TBDex.
+4. Test the updated vectors against at least two implementations of TBDex.
 5. Once you are confident that your test vector works as desired, open a PR against `tbdex`.
--- a/hosted/test-vectors/protocol/README.md
+++ b/hosted/test-vectors/protocol/README.md
@@ -21,7 +21,7 @@ Each vector file must follow the schema in `vectors.json.schema`.
 5. Once you are confident that your test vector works as desired, open a PR against `tbdex`.
 
 ### Updating Existing Test Vectors
-You should update an existing test vector if and only if there is a backwards incompatible spec change. This should be rare. If this is not the case, you should generate a new test vector that captures the desired scenario. If you must update an existing test vector, follow similar steps to generating a new test vector.
+You should update an existing test vector if and only if there is a backwards incompatible spec change that must be reflected in the test vector. This should be rare. If this is not the case, you should generate a new test vector that captures the desired scenario. If you must update an existing test vector, follow similar steps to generating a new test vector.
 1. Use an implementation of TBDex that supports the desured scenario to create the `input`, `output`, and `error`.
 2. Validate the test vector structure.
   a. Set up with `cd hosted/test-vectors/protocol & npm install`.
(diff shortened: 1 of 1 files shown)


mistermoe (approved):
@diehuxx does this apply to updating existing test vectors as well? asking bc of step 1 in this list. either way ➕

KendallWeihe (approved):

hosted/test-vectors/protocol/README.md
```diff
@@ -11,11 +11,21 @@ Each vector file must follow the schema in `vectors.json.schema`.
...
   a. Set up with `cd hosted/test-vectors/protocol & npm install`.
   b. Run with `npm run validate-vectors`.
-4. Test the updated vectors against an implmentation of TBDex.
-5. Once you are confident that your test vector works as desired, open a PR against `tbdex`.
\ No newline at end of file
+4. Test the generated vectors against at least two implementations of TBDex.
+5. Once you are confident that your test vector works as desired, open a PR against `tbdex`.
+
+### Updating Existing Test Vectors
+You should update an existing test vector if and only if there is a backwards incompatible spec change. This should be rare. If this is not the case, you should generate a new test vector that captures the desired scenario. If you must update an existing test vector, follow similar steps to generating a new test vector.
```
diehuxx, line 24:
```suggestion
You should update an existing test vector if and only if there is a backwards incompatible spec change that must be reflected in the test vector. This should be rare. If this is not the case, you should generate a new test vector that captures the desired scenario. If you must update an existing test vector, follow similar steps to generating a new test vector.
```

hosted/test-vectors/protocol/README.md
```diff
@@ -11,11 +11,21 @@ Each vector file must follow the schema in `vectors.json.schema`.
...
+### Generating New Test Vectors
 1. Create a new test vector file in `hosted/test-vectors/protocol/vectors/`
 2. Use an implementation of TBDex that supports the desired scenario to create the `input`, `output`, and `error`.
 3. Validate the test vector structure.
   a. Set up with `cd hosted/test-vectors/protocol & npm install`.
   b. Run with `npm run validate-vectors`.
-4. Test the updated vectors against an implmentation of TBDex.
-5. Once you are confident that your test vector works as desired, open a PR against `tbdex`.
\ No newline at end of file
+4. Test the generated vectors against at least two implementations of TBDex.
```
KendallWeihe, line 20: only probs with this is the test vector (being _new_) may introduce an incompatibility with all existing implementations

nevertheless, still good advice

hosted/test-vectors/protocol/README.md
```diff
@@ -11,11 +11,21 @@ Each vector file must follow the schema in `vectors.json.schema`.
...
+5. Once you are confident that your test vector works as desired, open a PR against `tbdex`.
+
+### Updating Existing Test Vectors
+You should update an existing test vector if and only if there is a backwards incompatible spec change that must be reflected in the test vector. This should be rare. If this is not the case, you should generate a new test vector that captures the desired scenario. If you must update an existing test vector, follow similar steps to generating a new test vector.
+1. Use an implementation of TBDex that supports the desured scenario to create the `input`, `output`, and `error`.
```
diehuxx, line 25:
```suggestion
1. Use an implementation of TBDex that supports the desired scenario to create the `input`, `output`, and `error`.
```

hosted/test-vectors/protocol/README.md
diehuxx: From the notes below
> You should update an existing test vector if and only if there is a backwards incompatible spec change that must be reflected in the test vector.

If there is a backwards incompatibility, we should be updating existing vectors, not adding entirely new ones. If we add a new test vector, the TBDex implementations that don't support the new scenario can omit the new test vector from their test suite until they support that scenario. The reason we require two implementations to support a vector before adding/updating it is we need to ensure the correctness of the test vector itself.

hosted/test-vectors/protocol/README.md
KendallWeihe: always be catching typos 👏

hosted/test-vectors/protocol/README.md
KendallWeihe: Oops, I read that and somehow interpreted it backwards, my bad

Okay so if there is an incompatibility induced with the given change then we should update the existing test vector, else if we don't induce an incompatibility then we should create a new test vector? 

Do we version test vectors alongside of the given tbDEX version? Or, we haven't got that far have we?

PR #247, opened 2024-02-22
regenerated test vector to add a valid DID for to field in `Message.metadata.from|to`

jiyoontbd:


hosted/test-vectors/protocol/vectors/parse-close.json
```diff
@@ -1,19 +1,19 @@
...
-  "input": "{\"metadata\":{\"from\":\"did:dht:abecscpi377h647nj5j8oz6yq544i8bzdze83mxsghskahq35ddy\",\"to\":\"did:ex:alice\",\"exchangeId\":\"abcd123\",\"kind\":\"close\",\"id\":\"close_01hpsc6aqten79yhw1gg25prmj\",\"createdAt\":\"2024-02-16T16:32:31.482Z\"},\"data\":{\"reason\":\"The reason for closing the exchange\"},\"signature\":\"eyJhbGciOiJFZERTQSIsImtpZCI6ImRpZDpkaHQ6YWJlY3NjcGkzNzdoNjQ3bmo1ajhvejZ5cTU0NGk4YnpkemU4M214c2doc2thaHEzNWRkeSMwIn0..LfYrchKuy0KtdRl-HRgLDSwRvzVAz6mApuQdJU5prOchLMRJkzcjROWxn5JbzkwVajH6VN4Xmhv6UcOfmZmoCA\"}",
+  "input": "{\"metadata\":{\"from\":\"did:dht:5feinba343k9fhy4zgdcp64bm6y7si39ru1pfjmrbxw9tkggbeuo\",\"to\":\"did:dht:t13oezyg87588ja11zoox6swtt7a8gq4zbmc6su5xbjy8xgdb46o\",\"exchangeId\":\"rfq_01hq9j1a70f3zrrh11n3q9x3jk\",\"kind\":\"close\",\"id\":\"close_01hq9j1a71fgh9qszsx06mefyn\",\"createdAt\":\"2024-02-22T23:22:29.473Z\"},\"data\":{\"reason\":\"The reason for closing the exchange\"},\"signature\":\"eyJhbGciOiJFZERTQSIsImtpZCI6ImRpZDpkaHQ6NWZlaW5iYTM0M2s5Zmh5NHpnZGNwNjRibTZ5N3NpMzlydTFwZmptcmJ4dzl0a2dnYmV1byMwIn0..Nnn8v5RZIgkA8rc9TC9WgKqdx1xe9F8197vhLqmlMQxv-U62jbORV5vNJXdzPlOo6PfgK84OdAAx1cuhsvcZCw\"}",
```
jiyoontbd, line 3: all these were regenerated because Message.metadata.to field was `did:ex:alice` and that fails our schema validation - it must be a real DID

KendallWeihe:
Could you add more description as to what made the pre-existing DIDs invalid? (or maybe point to a ticket which describes the issue)

I'm cool with the CODEOWNERs change, though it's worthy of a discussion (changes here being so upstream of everything else makes changes here have leverage, resulting in large downstream implications; my point being, if there ever were a need to create a constitution of governance then it should be had here -- "we're not there yet" is a valid response such that we can iterate quicker right now). Which, there may have been a discussion for which I wasn't privy to, but if not, could we make the CODEOWNERs a dedicated PR and/or ticket on the backlog?

jiyoontbd:
> Could you add more description as to what made the pre-existing DIDs invalid? (or maybe point to a ticket which describes the issue)

yes, can do! so i found this while getting my tbdex-js pr up to speed with main, and found that tbdex had some schema changes that enforced `Message.metadata.from/to` to be a real DID. the test vector was not updated with this change, and some of the test vectors had the value for those fields set to an invalid DID like `did:ex:alice` or `did:ex:pfi`. tests in tbdex-js and tbdex-kt that used the test vectors were failing because of this

> I'm cool with the CODEOWNERs change, though it's worthy of a discussion (changes here being so upstream of everything else makes changes here have leverage, resulting in large downstream implications; my point being, if there ever were a need to create a constitution of governance then it should be had here -- "we're not there yet" is a valid response such that we can iterate quicker right now). Which, there may have been a discussion for which I wasn't privy to, but if not, could we make the CODEOWNERs a dedicated PR and/or ticket on the backlog?

happy to move the codeowners change out to a separate PR!

hosted/json-schemas/message.schema.json
```diff
@@ -11,7 +11,7 @@
           "description": "The sender's DID"
         },
         "to": {
-          "$ref": "https://tbdex.dev/definitions.json#/definitions/did",
+          "$ref": "definitions.json#/definitions/did",
```
jiyoontbd, line 14: changed to be consistent with how we use $ref elsewhere https://github.com/TBD54566975/tbdex/pull/247/files#diff-781d77c475148a35130d81220b3a6d100357fc53576806be6dda9a528f1ae283L10

KendallWeihe:
> the test vectors had the value for those fields set to an invalid DID like `did:ex:alice` or `did:ex:pfi`

Got it, thanks! A good amount of these seem like `did:dht`-to-`did:jwk` changes. I would imagine we want `did:dht` in some of our test vectors, albeit maybe not currently? Or maybe it's beside the point and either or are fine and the changes here were just a byproduct of you generating new test vectors. Just raising awareness cc @phoebe-lew

jiyoontbd:
> > the test vectors had the value for those fields set to an invalid DID like `did:ex:alice` or `did:ex:pfi`
> 
> Got it, thanks! A good amount of these seem like `did:dht`-to-`did:jwk` changes. I would imagine we want `did:dht` in some of our test vectors, albeit maybe not currently? Or maybe it's beside the point and either or are fine and the changes here were just a byproduct of you generating new test vectors. Just raising awareness cc @phoebe-lew

yes, so i actually did initially opt to use did:dht, but found an issue while using did:dht:... for DIDs in test vectors and then ran tests in tbdex-kt and ran into an issue. https://github.com/TBD54566975/tbdex-kt/issues/180

i should probably write an issue to revert the DIDs back to did:dht once that issue is resolved. wrote issue here: https://github.com/TBD54566975/tbdex/issues/249

diehuxx (approved):

jiyoontbd:
So would that mean we have to first address the did:dht issue i found while
running test vector tests in tbdex-kt? I only used did:jwk to unblock
myself so all the tests would pass with the latest test vectors.

mistermoe:
@jiyoontbd just fixed the issue you found earlier! 

Related issue:  https://github.com/TBD54566975/tbdex-kt/issues/180
PR: https://github.com/TBD54566975/web5-kt/pull/251

diehuxx (approved):

KendallWeihe (approved):

PR #283, opened 2024-03-21
Wrap list offerings response in data object

diehuxx:
All other endpoints that return data follow the pattern of returning items inside `{ data: <stuff> }`.

### DIFF
--- a/specs/http-api/README.md
+++ b/specs/http-api/README.md
@@ -284,10 +284,10 @@ False
 
 
 ### Response
-| Status             | Body                  |
-| ------------------ | --------------------- |
-| `200: OK`          | `Offering[]`          |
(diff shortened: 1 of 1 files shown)


phoebe-lew (approved):

KendallWeihe:
> does this need to trickle through to implementations?

[tbdex-js is good to go](https://github.com/TBD54566975/tbdex-js/blob/74b6aded68f4aaa5103abbfde74138fb2e62eab2/packages/http-server/src/request-handlers/get-offerings.ts#L28)

diehuxx:
[tbdex-swift is good to go ](https://github.com/TBD54566975/tbdex-swift/blob/main/Sources/tbDEX/HttpClient/tbDEXHttpClient.swift#L215)

Issue #242, opened 2024-02-20
Distinguish between success and failure in `Close` message


phoebe-lew:
> can we provide a boolean property on the Close message indicating if the transaction was completed successfully or not. As a wallet app, all I get is Close and some String reason (again that could be in another language). I may want to add conditional logic to my app to display certain flows based on if the transaction completed successfully or not, and right now that's not possible without trying to programmatically decipher the Close.reason and determine if this String means things are good or not. 

from @angiejones

diehuxx:
I would argue a `Close` is always "bad". To know if an exchange is successful we use `OrderStatus`, not `Close`. Currently, in `tbdex-js`, `tbdex-kt`, and the spec, an exchange is terminal either with `Close` or `OrderStatus` where `Close` indicates a premature termination. In the existing implementations, after an `OrderStatus`, you actually cannot add a `Close` message.

From [the spec section for `Close`](https://github.com/TBD54566975/tbdex/tree/9a9ed69/specs/protocol#close):
> > Alice -> PFI: "Not interested anymore." or "oops sent by accident"
>
> > PFI -> Alice: "Can't fulfill what you sent me for whatever reason (e.g. RFQ is erroneous, don't have enough liquidity etc.)" or "Your exchange is completed"
>
> a Close can be sent by Alice or the PFI as a reply to an RFQ or a Quote. It indicates a terminal state. No messages can be added to an exchange after a Close.

angiejones:
> I would argue a `Close` is always "bad". To know if an exchange is successful we use `OrderStatus`, not `Close`. Currently, in `tbdex-js`, `tbdex-kt`, and the spec, an exchange is terminal either with `Close` or `OrderStatus` where `Close` indicates a premature termination. In the existing implementations, after an `OrderStatus`, you actually cannot add a `Close` message.
> 
> From [the spec section for `Close`](https://github.com/TBD54566975/tbdex/tree/9a9ed69/specs/protocol#close):
> 
> > > Alice -> PFI: "Not interested anymore." or "oops sent by accident"
> > 
> > 
> > > PFI -> Alice: "Can't fulfill what you sent me for whatever reason (e.g. RFQ is erroneous, don't have enough liquidity etc.)" or "Your exchange is completed"
> > 
> > 
> > a Close can be sent by Alice or the PFI as a reply to an RFQ or a Quote. It indicates a terminal state. No messages can be added to an exchange after a Close.

[...]

PR #286, opened 2024-03-22
Add Close.success to the protocol spec

### LINKED ISSUE #257: New rules for `Close` messages
Intentionally creating a bit of a duplicate from https://github.com/TBD54566975/tbdex/issues/242 with the hope of focusing aim and consolidating consensus requirements

1. `Close` messages are valid after any message type 
2. `Close` message introduces a new optional property `success` of type boolean

The logic of `Close.success` is as such:

```
if Close.success == null then that means Close was sent prior-to the Order execution
else if Close.success == true then Close was sent after a successful Order execution
else if Close.success == false then Close was sent after a failed Order execution
```

## `Close` after `Order`

`Close` messages _may_ be sent by Alice to the PFI during an actively executing order (AKA a state after, but not necessarily immediately after, an `Order` but prior to a `Close` where the `Close.success` is `true`), but that doesn't necessarily _guarantee_ a successful canceled execution. 

In the event Alice sends a `Close` message during an actively executing order:

```
if Close.success == true then Alice failed to cancel the executing order
else if Close.success == false then Alice successfully cancelled the executing order
```

---

1. We need to update the relevant spec docs here in this repo
2. We need to trickle through the changes through to various implementations

KendallWeihe:
Closes #257

Reference https://github.com/TBD54566975/tbdex/issues/242#issuecomment-1977002689

1. Close messages are valid after any message type
2. Close message introduces a new optional property `success` of type boolean

---

# Implementors

@jiyoontbd @kirahsapong @diehuxx (please tag anyone else in the comment section if appropriate)

Two things.

First, `Close` is now valid after any message type. I believe in some implementations we have a function akin to `isValidNext()` which inspects the message type (& its contents?) and determines if the given message is valid as the next message in the exchange. In places where we do have this code, we'll need to ensure it allows `Close` messages at any point.

Second, add a new property `Close.success` which is an _optional_ boolean. Ensure that `success` is accessible during creation/validation/parsing/etc (wherever appropriate).

### DIFF
--- a/specs/protocol/README.md
+++ b/specs/protocol/README.md
@@ -496,14 +496,14 @@ Currency amounts have type `DecimalString`, which is string containing a decimal
 
 > PFI -> Alice: "Can't fulfill what you sent me for whatever reason (e.g. RFQ is erroneous, don't have enough liquidity etc.)" or "Your exchange is completed"
 
-a `Close` can be sent by Alice _or_ the PFI as a reply to an RFQ or a Quote. It indicates a terminal state. No messages can be added to an exchange after a `Close`.
+A `Close` indicates a terminal state; no messages are valid after a `Close`. 
 
-| Field    | Data Type | Required | Description                                                  |
-| -------- | --------- | -------- | ------------------------------------------------------------ |
-| `reason` | string    | N        | an explanation of why the exchange is being closed/completed |
+| Field     | Data Type | Required | Description                                                  |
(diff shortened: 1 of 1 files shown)


jiyoontbd (approved):
✨

diehuxx (approved):

phoebe-lew (approved):

PR #291, opened 2024-03-25
revert external refs to types

kirahsapong:
external $refs introduces ajv generating a `require` in our ts validator. these $refs are valid to use but due to ajv injecting require syntax with externally resolved urls, reverting to `type: object`


jiyoontbd (approved):

decentralgabe:
why is require syntax a problem? if we are requiring json schemas it would be better to enforce it

decentralgabe:
is there a workaround that can maintain the $ref?

the value of json schema validation is to enforce the shape we require...without being able to enforce that shape (and just a generic `object`) is the schema giving us much utility?

phrased differently - if we can't enforce the properties are valid json schemas here, where else can we do it?

Issue #274, opened 2024-03-14
Specify structure of private fields in RFQ


diehuxx:
## History
Previously, we considered the concept of a top-level private property in RFQ, which would contain the private data specified by the Offering, where corresponding Rfq.data would be hashed. This has proved to be complicated in design and implementation.

We opened a PR #232 with a proposal, however @decentralgabe pointed out that 1) we should discuss in an Issue before opening a PR, and 2) the proposal partially implements SD-JWT and we might be better off just using SD-JWT.

diehuxx:
@mistermoe @decentralgabe and I chatted offline. Looks like we can settle on the following design. PII fields (`claims`, `payin.paymentDetails`, and `payout.paymentDetails`) are not signed over and their salted hashes are signed over. So I'll update #232 to specify salted hashes.

jiyoontbd:
just want to clarify - is this work slated for 1.0?

diehuxx:
Hi @jiyoontbd -- good question, yes it is. I just added labels/milestones to this issue to denote that it's part of 1.0. Since we're coming up on the deadline, this ticket is my main priority this week. I'm planning to open PRs for the spec and JS SDK today.

PR #294, opened 2024-03-28
#274 Create salted hashes of privateData in Rfq

diehuxx:
## Context
#274

Previously, we considered the concept of a top-level unstructured `private` property in RFQ, which would optionally contain the private data specified by the Offering, where corresponding Rfq.data would be hashed. This PR lays out a simplified design compared to what we had before.

It is useful to be able to detach private fields from the RFQ payload and still be able to verify message integrity. So we create a separate section for private data.

## Q&A
1. Why call it `privateData` when we used to call it `private`?
Since `private` is a reserved word in most programming languages, it was awkward to call the field `private` in some places and have to think of another name in other places.

2. Why not use SD-JWT?
SD-JWT is more complex than our use case requires. There are only a few fields that we need to selectively disclose in an RFQ. The design laid out in this PR takes some ideas from SD-JWT without needing to implement the rest of the kitchen sink.

## Test vectors
This PR updates the existing `parse-rfq` test vectors because the new design constitutes a breaking change to RFQ message structure. This PR also adds a net new test vector related to RFQs called `parse-rfq-omit-private-data`.

TODO: Support updated test vectors in at least two tbdex implementations
- [x] tbdex-js [PR](https://github.com/TBD54566975/tbdex-js/pull/215)
- [ ] tbdex-kt
- [ ] tbdex-swift

### DIFF
--- a/hosted/json-schemas/rfq.schema.json
+++ b/hosted/json-schemas/rfq.schema.json
@@ -8,12 +8,12 @@
       "type": "string",
       "description": "Offering which Alice would like to get a quote for"
     },
-    "claims": {
+    "claimsHashes": {
       "type": "array",
       "items": {
         "type": "string"
       },
-      "description": "Presentation Submission that fulfills the requirements included in the respective Offering"
+      "description": "Digests of Presentation Submissions that fulfills the requirements included in the respective Offering"
     },
     "payin": {
       "type": "object",
@@ -25,9 +25,9 @@
           "type": "string",
           "description": "Type of payment method e.g. BTC_ADDRESS, DEBIT_CARD, MOMO_MPESA"
         },
-        "paymentDetails": {
-          "type": "object",
-          "description": "An object containing the properties defined in the respective Offering's requiredPaymentDetails json schema"
+        "paymentDetailsHash": {
+          "type": "string",
+          "description": "Digest of an object containing the properties defined in the respective Offering's requiredPaymentDetails json schema"
         }
       },
       "required": ["amount", "kind"]
@@ -39,13 +39,13 @@
           "type": "string",
           "description": "Selected payout method from the respective offering"
         },
-        "paymentDetails": {
-          "type": "object",
-          "description": "An object containing the properties defined in the respective Offering's requiredPaymentDetails json schema"
+        "paymentDetailsHash": {
+          "type": "string",
+          "description": "Digest of an object containing the properties defined in the respective Offering's requiredPaymentDetails json schema"
         }
       },
       "required": ["kind"]
     }
   },
-  "required": ["offeringId", "claims", "payin", "payout"]
+  "required": ["offeringId", "claimsHashes", "payin", "payout"]
 }
--- a/hosted/json-schemas/rfq.schema.json
+++ b/hosted/json-schemas/rfq.schema.json
@@ -8,11 +8,8 @@
       "type": "string",
       "description": "Offering which Alice would like to get a quote for"
     },
-    "claimsHashes": {
-      "type": "array",
-      "items": {
-        "type": "string"
-      },
+    "claimsHash": {
+      "type": "string",
       "description": "Digests of Presentation Submissions that fulfills the requirements included in the respective Offering"
     },
     "payin": {
@@ -47,5 +44,5 @@
       "required": ["kind"]
     }
   },
-  "required": ["offeringId", "claimsHashes", "payin", "payout"]
+  "required": ["offeringId", "payin", "payout"]
 }
--- a/specs/protocol/README.md
+++ b/specs/protocol/README.md
@@ -341,7 +341,7 @@ The `metadata` object contains fields _about_ the message and is present in _eve
 
 
 | Field        | Required (Y/N) | Description                                                                                                                                                                                                                                                                                                    |
-|--------------|----------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| ------------ | -------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `from`       | Y              | The sender's DID                                                                                                                                                                                                                                                                                               |
 | `to`         | Y              | the recipient's DID                                                                                                                                                                                                                                                                                            |
 | `kind`       | Y              | e.g. `rfq`, `quote` etc. This defines the `data` property's _type_                                                                                                                                                                                                                                             |
@@ -408,7 +408,7 @@ Currency amounts have type `DecimalString`, which is string containing a decimal
 > Alice -> PFI: "OK, that offering looks good. Give me a Quote against that Offering, and here is how much USD (payin currency) I want to trade for BTC (payout currency). Here are the credentials you're asking for, the payment method I intend to pay you USD with, and the payment method I expect you to pay me BTC in."
 
 | field        | data type                                        | required | description                                                 |
-|--------------|--------------------------------------------------|----------|-------------------------------------------------------------|
+| ------------ | ------------------------------------------------ | -------- | ----------------------------------------------------------- |
 | `offeringId` | string                                           | Y        | Offering which Alice would like to get a quote for          |
 | `claimsHash` | string[]                                         | Y        | Salted hash of the claims appearing in `privateDate.claims` |
 | `payin`      | [`SelectedPayinMethod`](#selectedpaymentmethod)  | Y        | selected payin amount, method, and details                  |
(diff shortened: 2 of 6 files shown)


decentralgabe (approved):
one thing I realize, is that there is an assumption baked into this proposal that the claims included in the presentation submission are issued in a format that's not selectively disclosable. if, for example, we accepted claims that were SD-JWT we would run into a problem where we would need to submit disclosures of the SD-JWT in multiple places. 

it seems like we should decouple the requirements of the submission from its credential format.

this could look like include a new type of [claim format](https://identity.foundation/claim-format-registry/) so that the concern of disclosing PII is handled at the presentation layer instead of the protocol layer in a `privateData` field.

I understand time is of the essence, so I am OK with the proposal as-written even though I do have some hesitations around it.  May be good to create a follow-up issue to continue this discussion and track improvements, like what we should do when we do adopt SD-JWT or other credential formats.

jiyoontbd (approved):

hosted/json-schemas/rfq.schema.json
```diff
@@ -8,12 +8,12 @@
...
       "description": "Offering which Alice would like to get a quote for"
     },
-    "claims": {
+    "claimsHashes": {
       "type": "array",
       "items": {
         "type": "string"
       },
-      "description": "Presentation Submission that fulfills the requirements included in the respective Offering"
+      "description": "Digests of Presentation Submissions that fulfills the requirements included in the respective Offering"
```
KendallWeihe, line 16: Do you think it would be good to be slightly more specific (also in the `paymentDetails` in the lines below)

```suggestion
      "description": "Salted hash of Presentation Submissions that fulfills the requirements included in the respective Offering"
```

KendallWeihe: Oh yeah, we have a dedicated section on Digest's so idk maybe not? What do you think @diehuxx? My concern is some places we have language around "digest" (which is more broad) and other places "salted hash" (which is a particular type of digest) and how that may lead to ambiguity and/or confusion.

specs/protocol/README.md
```diff
@@ -422,25 +419,38 @@ Currency amounts have type `DecimalString`, which is string containing a decimal
...
-| `claims`     | string[]                                         | Y        | an array of claims that fulfill the requirements declared in an [Offering](#offering) |
-| `payin`      | [`SelectedPayinMethod`](#selectedpaymentmethod)  | Y        | selected payin amount, method, and details                                            |
-| `payout`     | [`SelectedPayoutMethod`](#selectedpaymentmethod) | Y        | selected payout method, and details                                                   |
+| field          | data type                                        | required | description                                                                |
+|----------------|--------------------------------------------------|----------|----------------------------------------------------------------------------|
+| `offeringId`   | string                                           | Y        | Offering which Alice would like to get a quote for                         |
+| `claimsHashes` | string[]                                         | Y        | an array of salted digests of the claims appearing in `privateDate.claims` |
```
KendallWeihe, line 425:
```suggestion
| `claimsHashes` | string[]                                         | Y        | an array of salted hashes of the claims appearing in `privateDate.claims` |
```

specs/protocol/README.md
```diff
@@ -422,25 +419,38 @@ Currency amounts have type `DecimalString`, which is string containing a decimal
...
+|----------------------|-----------------------------------|----------|-----------------------------------------------------------------------------|
+| `amount`             | [`DecimalString`](#decimalstring) | Y        | Amount of payin currency you want in exchange for payout currency           |
+| `kind`               | string                            | Y        | Type of payment method (i.e. `DEBIT_CARD`, `BITCOIN_ADDRESS`, `SQUARE_PAY`) |
+| `paymentDetailsHash` | string                            | N        | A salted digest of `privateData.payin.paymentDetails`                       |
 
 #### `SelectedPayoutMethod`
+| field                | data type | required | description                                                                 |
+|----------------------|-----------|----------|-----------------------------------------------------------------------------|
+| `kind`               | string    | Y        | Type of payment method (i.e. `DEBIT_CARD`, `BITCOIN_ADDRESS`, `SQUARE_PAY`) |
+| `paymentDetailsHash` | object    | N        | A salted digest of `privateData.payout.paymentDetails`                       |
```
KendallWeihe, line 440:
```suggestion
| `paymentDetailsHash` | object    | N        | A salted hash of `privateData.payout.paymentDetails`                       |
```

"salted digest" is technically correct (I think?) but we use "salted hash" elsewhere so consistency, and my understanding is "salted hash" is the common term to refer to this data type

so anyways, not opposed to "salted digest" but just calling it out because I see it

KendallWeihe (approved):
some optional comments but lgtm ✅

mistermoe:
@diehuxx just opened a PR that includes a handful of recommended extensions that make writing markdown subjectively easier (e.g. auto formatting tables, managing table of contents for you) https://github.com/TBD54566975/tbdex/pull/295. can you use [Markdown All in One](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one) and [Markdown Table](https://marketplace.visualstudio.com/items?itemName=TakumiI.markdowntable) to fix the ToC and table formatting?

specs/protocol/README.md
```diff
@@ -362,37 +362,34 @@ The `metadata` object contains fields _about_ the message and is present in _eve
...
 
-The `private` field is ephemeral and **MUST** only be present when the message is initially sent to the intended recipient
+Message kinds other than RFQ may NOT have property `privateData`. The [RFQ `privateData` section](#rfq-request-for-quote) specifies the content of RFQ `data` and `privateData`.
 
-The value of `private` **MUST** be a JSON object that matches the structure of `data`. The properties present within `private` **MUST** only be the properties of `data` that include the hash counterpart.
-
-> **Note**
-> Rationale behind the `private` JSON object matching the structure of `data` is to simplify programmatic hash evaluation using JSONPath to pluck the respective hash from `data`. **NOTE**: we should try this to make sure it's actually "easy"
+The `privateData` field is ephemeral and **MUST** only be present when the message is initially sent to the intended recipient
```
mistermoe, line 372: i think we need to explain the `*Hashes` properties in this section. it's currently only explained in an `e.g.`

hosted/json-schemas/rfq.schema.json
diehuxx: How do you feel about "salted digest"? It sounds much less appetizing than a salted hash, but may be more specific to what we actually mean. I waffled on this one tbh.

specs/protocol/README.md
diehuxx: Ah, this answers [my question above](https://github.com/TBD54566975/tbdex/pull/294#discussion_r1543383343) about whether you like the term "salted digest".

specs/protocol/README.md
diehuxx: Added a more explicit section for `*Hashes`

specs/protocol/README.md
diehuxx: Updated to "salted hash"

specs/protocol/README.md
```diff
@@ -422,25 +419,38 @@ Currency amounts have type `DecimalString`, which is string containing a decimal
...
-| `kind`           | string                            | Y        | Type of payment method (i.e. `DEBIT_CARD`, `BITCOIN_ADDRESS`, `SQUARE_PAY`)                       |
-| `paymentDetails` | object                            | N        | An object containing the properties defined in an Offering's `requiredPaymentDetails` json schema |
+| field                | data type                         | required | description                                                                 |
+|----------------------|-----------------------------------|----------|-----------------------------------------------------------------------------|
+| `amount`             | [`DecimalString`](#decimalstring) | Y        | Amount of payin currency you want in exchange for payout currency           |
+| `kind`               | string                            | Y        | Type of payment method (i.e. `DEBIT_CARD`, `BITCOIN_ADDRESS`, `SQUARE_PAY`) |
+| `paymentDetailsHash` | string                            | N        | A salted digest of `privateData.payin.paymentDetails`                       |
```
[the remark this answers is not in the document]
diehuxx: Updated to "salted hash"

specs/protocol/README.md
diehuxx: Updated to "salted hash"

specs/protocol/README.md
mistermoe: nice. reads really well! looks like there's a table formatting issue 

image

mistermoe:
what's the rationale behind `claimsHashes` over just `claimsHash`?

specs/protocol/README.md
```diff
@@ -354,37 +356,28 @@ The `metadata` object contains fields _about_ the message and is present in _eve
...
+    "payout": {
       "kind": "MOMO_MPESA",
-      "paymentDetails": <OBJ_HASH> <---- hash
+      "paymentDetailsHash": <OBJ_HASH> <---- hash
     }
   },
-  "private": {
-    "claims": <PRESENTATION_SUBMISSION>, <---- actual
+  "privateData": {
+    "claims": [<PRESENTATION_SUBMISSION>], <---- actual
```
mistermoe, line 380: the claims array is an array of VCs not presentation submission

diehuxx:
> what's the rationale behind claimsHashes over just claimsHash?

I was thinking that in subsequent minor versions, we could add more nuance in validating each of the claims hashes to allow for a subset claims to be disclosed in `privateData`. I'm not married to this idea, since I don't have a concrete use case in mind that requires it. If you think it's not worth the added complexity, I'm open to changing it.

mistermoe:
@diehuxx i think we'll probably want to lean in `sd-jwt` as much as possible wrt selective disclosure thangs. i think we should remove it.

mistermoe (approved):

PR #313, opened 2024-04-10
Specify null case for private data hashes

diehuxx:
Update spec for Rfq.data and Rfq.privateData to make `Rfq.data.claimsHash` optional. `claimsHash` is already optional on [tbdex-js](https://github.com/TBD54566975/tbdex-js/blob/main/packages/protocol/src/types.ts#L266), [tbdex-kt](https://github.com/TBD54566975/tbdex-kt/blob/main/protocol/src/main/kotlin/tbdex/sdk/protocol/models/MessageData.kt#L25), [tbdex-swift](https://github.com/TBD54566975/tbdex-swift/blob/main/Sources/tbDEX/Protocol/Models/Messages/RFQ.swift#L97), and [tbdex-dart](https://github.com/TBD54566975/tbdex-dart/blob/main/lib/src/protocol/models/message_data.dart#L92). This PR brings the spec into alignment with the intention and existing implementations.

Also, add spec text clarifying that expected behavior when `Rfq.data.*Hash` is null/omitted, then the corresponding `Rfq.privateData.*` field must also be null/omitted. 

This PR contains a change to JSON schema:
- [x] At least two implementations must support the new JSON schema. [tbdex-js](https://github.com/TBD54566975/tbdex-js/pull/243) and [tbdex-kt](https://github.com/TBD54566975/tbdex-kt/pull/239)


specs/protocol/README.md
```diff
@@ -408,19 +408,19 @@ Currency amounts have type `DecimalString`, which is string containing a decimal
...
+|--------------|--------------------------------------------------|----------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `offeringId` | string                                           | Y        | Offering which Alice would like to get a quote for                                                                                                                                                                                                                    |
+| `claimsHash` | string                                           | N        | Salted hash of the list the claims appearing in `privateData.claims`. Omit `claimsHash` to denote that the RFQ contains no claims, i.e. `privateData.claims` is omitted. If `privateData.claims` is an empty list, `claimsHash` is the salted hash of the empty list. |
```
decentralgabe, line 414: why have a salted hash of a list as opposed to no value at all?

diehuxx: I mention this in the description for `privateData.claims` that the RFQ creator should prefer to have null/no value rather than an empty list. But i also wanted to be explicit about how the empty list case should be handled.

mistermoe (approved):

specs/protocol/README.md
decentralgabe: I would remove the possibility altogether

diehuxx: Good call. Updated

specs/protocol/README.md
```diff
@@ -435,23 +435,23 @@ In order to prevent storing this sensitive data with the message itself, an RFQ
...
+|----------|---------------------------------------------------|----------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `salt`   | string                                            | Y        | Randomly generated salt used for hashing PrivateData fields                                                                                                                                                                                                                                            |
+| `claims` | string[]                                          | N        | an array of claims that fulfill the requirements declared in an [Offering](#offering). If the creator of the RFQ wants to include no claims, then `claims` SHOULD be omitted rather than being an empty list. `claims` must not be an empty list. |
```
KendallWeihe, line 452:
```suggestion
| `claims` | string[]                                          | N        | an array of claims that fulfill the requirements declared in an [Offering](#offering). If the creator of the RFQ wants to include no claims, then `claims` MUST be omitted rather than being an empty list. `claims` must not be an empty list. |
```

right?

KendallWeihe: and you can remove the last bit as it's redundant 

```suggestion
| `claims` | string[]                                          | N        | an array of claims that fulfill the requirements declared in an [Offering](#offering). If the creator of the RFQ wants to include no claims, then `claims` MUST be omitted rather than being an empty list. |
```

KendallWeihe (approved):

Issue #312, opened 2024-04-09
Remove `queryParams` from `ListExchanges`


kirahsapong:
With the spec returning `{ data: [id, id, id] }` from any `GET /exchanges` call, it becomes redundant to pass queryParams of exchangeIds. Should remove

thehenrytsai:
This is now done.

PR #315, opened 2024-04-18
#312 - Removed query params for listing exchanges

thehenrytsai:
Query params of exchange IDs doesn't not make sense when are you are trying to fetch those IDs in the first place (see #312).


mistermoe (approved):
thanks @thehenrytsai !

jiyoontbd (approved):

PR #325, opened 2024-05-30
updating error response structure tables

### LINKED ISSUE #322: HTTP API spec: update error response structure
Our current error object in the [http api spec](https://github.com/TBD54566975/tbdex/tree/main/specs/http-api#error-responses) follows error object structure in [json:api spec](https://jsonapi.org/format/#error-objects). 

However, the set of fields inside each error object either a) doesn't apply well to the kind of details a PFI might want to return, or b) is not relevant at all.

Below is an example of the proposed new error response structure:
```json
{
  "message": "Missing field: payin.amount",
  "id": "9af2bf88-e4f4-4f81-8ba9-55eaeeb718e2",
  "code": "TBDEX_MESSAGE_VALIDATION_ERROR",
  "details": [{ 
    "message": "Payin amount must be present.",
    "path": "$.payin.amount" 
  }]
}
```

### Error response structure
| Field     | Type          | Required | Description                                                                            |
| --------- | ------------- | -------- | -------------------------------------------------------------------------------------- |
| `message` | String        | Y        | A human-readable explanation specific to this occurrence of the problem.               |
| `id`      | String        | N        | Optional server-generated request-specific ID, useful for diagnosing unexpected errors |
| `code`    | String        | Y        | An application-specific error code, expressed as a string value.                       |
| `details` | ErrorDetail[] | N        | Optional array of `ErrorDetail` objects                                                |

### ErrorDetail structure
| Field     | Type   | Required | Description                                                              |
| --------- | ------ | -------- | ------------------------------------------------------------------------ |
| `message` | String | N        | A human-readable explanation specific to this occurrence of the problem. |
| `path`    | String | N        | Path where validation failed (i.e. JSON schema path)                     |

The introduction of a required `code` field is significant - this implies that the spec would need to provide specific code strings and when a PFI would want to use it to form the error response body. 

The reason for `code` being a required field is that it would significantly reduce the amount of "error deduction" that wallet developers would have to do when parsing the error, as the top level `message` field (arbitrary string set by the PFI) + HTTP status code may not provide enough context. With the introduction of `code` field, the wallet developer can focus on displaying the appropriate error message depending on the known codes.

jiyoontbd:
closes #322

### DIFF
--- a/specs/http-api/README.md
+++ b/specs/http-api/README.md
@@ -17,7 +17,6 @@ Version: Draft
   - [Error response structure](#error-response-structure)
   - [Error `code`](#error-code)
   - [ErrorDetail structure](#errordetail-structure)
-  - [| `path`    | String | N        | Path where validation failed (i.e. JSON schema path)                     |](#-path-----string--n---------path-where-validation-failed-ie-json-schema-path---------------------)
   - [Example](#example-1)
 - [Exceptions](#exceptions)
 - [Query Params](#query-params)
@@ -120,6 +119,7 @@ If the serviceEndpoint is itself a DID, this DID should resolve to a document an
 | --------- | ------ | -------- | ------------------------------------------------------------------------ |
 | `message` | String | N        | A human-readable explanation specific to this occurrence of the problem. |
 | `path`    | String | N        | Path where validation failed (i.e. JSON schema path)                     |
+
 ---
 
 ## Example
--- a/specs/http-api/README.md
+++ b/specs/http-api/README.md
@@ -14,7 +14,10 @@ Version: Draft
 - [Discoverability](#discoverability)
   - [Example](#example)
 - [Error Responses](#error-responses)
-  - [Error object](#error-object)
+  - [Error response structure](#error-response-structure)
+  - [Error `code`](#error-code)
+  - [ErrorDetail structure](#errordetail-structure)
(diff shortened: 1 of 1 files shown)


specs/http-api/README.md
```diff
@@ -109,41 +101,38 @@ If the serviceEndpoint is itself a DID, this DID should resolve to a document an
...
+| --------- | ------------- | -------- | -------------------------------------------------------------------------------------- |
+| `message` | String        | Y        | A human-readable explanation specific to this occurrence of the problem.               |
+| `id`      | String        | N        | Optional server-generated request-specific ID, useful for diagnosing unexpected errors |
+| `code`    | String        | Y        | An application-specific error code, expressed as a string value.                       |
+| `details` | ErrorDetail[] | N        | Optional array of `ErrorDetail` objects                                                |
+
+## Error `code`
+| Code | Description |
+| ---- | ----------- |
+| TODO |
```
jiyoontbd, line 115: need input on the (exhaustive) list of codes to codify which code is used to describe which error scenario

KendallWeihe (approved):

specs/http-api/README.md
jiyoontbd: removing code for now

mistermoe (approved):

PR #341, opened 2024-07-03
adding cancel. modifying offering. clarifying purpose of close in protocol spec

### LINKED ISSUE #328: Consider introducing a `Cancel` message
# Context
Per EFTA regulation, consumers have the right to cancel a remittance transfer within 30 minutes after the _payin_ has been sent, so long as the transfer has not yet been picked up by the recipient or deposited into the recipient's account.

 If a consumer cancels a remittance transfer within the allowed time frame, the provider must refund the total amount of money paid by the consumer, including any fees, within three business days of receiving the cancellation request.

If we intend for tbDEX to be able to support traditional remittances we need to provide a means through which Alice can submit a cancelation

# Proposal
introduce a `Cancel` message with the following constraints:
* This message kind can _only_ be submitted by Alice 
* This message can _only_ be submitted _after_ an `Order` message and _before_ a `Close` message. in between `OrderStatus` messages is fine

## Message Data
The `data` property of a Cancel message would contain:

| Field    | Type   | Description            | Required (Y/N) |
| :------- | :----- | :--------------------- | :------------- |
| `reason` | string | reason for cancelation | N              |

## Cancel Submission
Sending a `Cancel` message could result in the following responses

### 202 - Accepted
The `Cancel` message has been accepted for _further processing_. 

> [!IMPORTANT]
> This does **not** imply a commitment to refund given the possibility that the payout has already been disbursed but the respective `OrderStatus` has not yet been added to the exchange

`OrderStatus` messages that follow after the `Cancel` message would be used to be provide further context on the status of the cancelation request e.g. 
* `REFUND_INITIATED` - The refund process has been initiated
* `CANCEL_DECLINED` - cancellation request has been declined. Payout amount has already been disbursed to the recipient.

### 400 - Bad Request
* Invalid Message
* Exchange has already been closed
* Cancelation has already been submitted
* Cancelation cannot be submitted given the current state of the exchange
* Cancelation Declined. (at the discretion of the PFI)

# Backwards Compatibility
Introducing this message kind would result in a version bump of `tbdex`. as such, submission of a `Cancel` message to any exchange with a `protocol` version less than the version that introduces `Cancel` would result in a `400: Bad Request`

# Other Considerations
We may want to consider adding an `cancelation` property to `Offering` that contains the following:

| Field       | Type    | Description                                                | Required (Y/N) |
| :---------- | :------ | :--------------------------------------------------------- | :------------- |
| `enabled`   | boolean | whether cancelation is enabled for the respective offering | Y              |
| `terms_url` | URI     | a link to a page that describes the terms of cancelation   | N              |
| `terms`     | string  | cancelation terms in plaintext                             | N              |

# Alternative Approaches
* provide a means through which cancellation can happen out-of-band. An example of this would be to allow for a response to an order message to include a `cancelUrl`.

jiyoontbd:
closes #328 

* changes offering schema to add the cancellation object
* introduces new cancel schema to add new cancel message
* changes parse offering test vector
* adds parse cancel test vector
* add angie in CODEOWNERS


nitro-neal (approved):
Same comment I had on the other pr, just want to get a path to make sure the vectors are in sync 👍

angiejones (approved):

hosted/json-schemas/offering.schema.json
```diff
@@ -142,12 +142,31 @@
...
+        "terms_url": {
+          "type": "string",
+          "description": "A link to a page that describes the terms of cancellation"
+        },
+        "terms": {
+          "type": "string",
+          "description": "A human-readable description of the terms of cancellation in plaintext"
+        }
+      },
+      "required": ["enabled"]
```
angiejones, line 162: Did we check with legal to see if cancelation terms are a requirement? According to Chad, it is

> Companies are required to provide the terms of cancellation under the EFTA (Electronic Fund Transfer Act) regulation. Specifically, the regulation mandates that remittance transfer providers must disclose certain key information to consumers, including the terms and conditions of cancellation rights.
> 
> The terms of cancellation must include:
> 
> * The right to cancel: Informing the consumer of their right to cancel the remittance transfer within 30 minutes.
> * Instructions on how to cancel: Providing clear and understandable instructions on how to execute the cancellation.
> * Conditions for cancellation: Stating that the right to cancel is applicable only if the transfer has not been picked up by the recipient or deposited into the recipient’s account.
>
> This information is usually provided in the pre-payment disclosures and the receipt given to the consumer at the time of the transaction. This ensures that consumers are fully aware of their rights and the procedures for exercising them.

jiyoontbd: ok, so we have either `terms` or `terms_url` - is what you're asking for to include either of those (or both) in a cancellation detail object?

jiyoontbd: created an issue for this. https://github.com/TBD54566975/tbdex/issues/349

PR #343, opened 2024-07-08
Fix privateData name

KendallWeihe:
This is called `privateData` everywhere but this one place (including the JSON schemas)


kirahsapong (approved):
nice find 🔥

PR #344, opened 2024-07-08
editing req body for create exchange and submit message

jiyoontbd:


specs/http-api/README.md
```diff
@@ -286,7 +276,7 @@ An exchange is a series of linked tbDEX messages between Alice and a PFI for a s
 Submits an RFQ (Request For Quote). Alice is asking the PFI to provide a Quote so she can evaluate it.
 
 ### Endpoint
-`POST /exchanges`
+`POST /exchanges/:exchange_id`
```
KendallWeihe, line 279: To be pedantic, to use an `:exchange_id` here then the method would need to be `PUT` because an id in the path means the resource is already created. So either `POST /exchanges` or `PUT /exchanges/:exchange_id` (which, the latter would conflict with the other endpoints we have). Open to pushback if there is some concrete reasoning I'm missing.

jiyoontbd: we've discussed this previously and landed here. https://github.com/TBD54566975/tbdex/issues/225

KendallWeihe (approved):

mistermoe (approved):

nitro-neal:
Is anything in the protocol updated? There have been a lot of changes and I want to make sure the test vectors are still synced up.

I have added this issue for quote which is out of sync - https://github.com/TBD54566975/tbdex/issues/345

Can we set up some time once all the new changes get in so we can re-sync the test vectors?

nitro-neal (approved):

PR #357, opened 2024-07-29
Make JSON schema URLs https

KendallWeihe:


jiyoontbd (approved):

nitro-neal (approved):

KendallWeihe:
@decentralgabe if there is any problem with this we can roll it back, it was blocking something pressing at the time

PR #364, opened 2024-08-07
Add formatting/linting recipes to `Justfile`

lamchau:
added 3 recipes to keep things consistent in the hosted repo
 - `clean`: remove temp files used during formatting
 - `format`: uses `jq` to format the files
 - `lint`: uses `jq` to verify all hosted `*.json` files

This also updates `schemas` which adds a verbose to see all the files that'll be copied to `.schemas` - while also adding a lint/verify step to ensure we have well-formed files.

Note: `format` depends on `lint` so any files that invalid JSON files will fail.


hosted/json-schemas/definitions.json
```diff
@@ -5,11 +5,11 @@
   "definitions": {
     "did": {
       "type": "string",
-      "pattern": "^did:([a-z0-9]+):((?:(?:[a-zA-Z0-9._-]|(?:%[0-9a-fA-F]{2}))*:)*((?:[a-zA-Z0-9._-]|(?:%[0-9a-fA-F]{2}))+))((;[a-zA-Z0-9_.:%-]+=[a-zA-Z0-9_.:%-]*)*)(\/[^#?]*)?([?][^#]*)?(#.*)?$"
```
lamchau, line 8: unfortunately `jq` turns this pattern into an invalid regex

Justfile
```diff
@@ -1,16 +1,43 @@
...
+    echo "[success] All JSON files are valid"
+
+schemas: lint
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    source_dir="hosted/json-schemas"
+    dest_dir=".schemas"
+
+    rm -rf $dest_dir
```
jiyoontbd, line 40: should we gitignore `.schemas/` if it's a temporary folder so we don't accidentally check it in?

jiyoontbd (approved):
🥳

Justfile
lamchau: yep! covered [here](https://github.com/TBD54566975/tbdex/blob/c645cc3860eb57f576a4ef716c9a7a5422860ea1/.gitignore#L5)! :)

hosted/json-schemas/definitions.json
lamchau: @jiyoontbd so i think as this PR currently stands, this current change _might_ break if a `did` has a path in it. how do we test these? i didn't see any test vectors with a _path_ so i wrote my own mini python test harness (hence the screenshotted error)

what do you recommend we do?

Justfile
```diff
@@ -1,16 +1,43 @@
...
+
+format: lint
+    #!/usr/bin/env bash
+    find hosted -type f -name "*.json" | while read file; do
+      jq --indent 2 . $file > $file.json.tmp
+      mv $file.json.tmp $file
+      printf "formatted: %s\n" $file
+    done
+
+lint: clean
```
KendallWeihe, line 18: should we add a github workflow which runs `just lint` such that PRs can't be merged if linting fails?

hosted/json-schemas/definitions.json
KendallWeihe: come to think of it, the `did` _definition_ probably ought to live under [the web5 repo](https://github.com/TBD54566975/web5-spec/), and within that repo we ought to have test vectors for DIDs like this. think we could get there in this PR? or we can add an issue and follow-up subsequently

Justfile
lamchau: ohh good idea!

i didn't spend too much time digging into this but i thought we generated the schemas somewhere so they'd always be valid? if these are manually crafted then that makes a ton of sense and i can definitely add that

hosted/json-schemas/definitions.json
lamchau: fixed a typo (ty @jiyoontbd) and added the escaped slash [123217d](https://github.com/TBD54566975/tbdex/pull/364/commits/123217dfe0a33f1d07732251f72289ddc450be9f). suppose we'll keep this (corrected) for now as the status quo.

@jiyoontbd chatted about this offline wrt to moving it to web5 so i think maybe it's a bigger discussion between the protocol and the SDK teams, but i'll let y'all figure out the best path forward here!

jiyoontbd: hmm, if we moved this to web5-spec, wouldn't each tbdex sdk now have to pull in that spec as a submodule just to be able to check if a DID is valid?

Justfile
lamchau: added a lint (on pull_request) and format (on push to `main`) [5e903a5](https://github.com/TBD54566975/tbdex/pull/364/commits/5e903a59832557bfb0a609e99263f5ef50408e80)

hosted/json-schemas/definitions.json
KendallWeihe: JSON schema supports remote references (`"$ref": "{url}"`) for definitions, so we would add the definition in web5-spec, deploy it and add the reference here. But it doesn't need to be a blocker on this PR, but at the least we should open an Issue to track it b/c I believe that's the ideal architectural solution.

.github/workflows/format.yml
```diff
@@ -0,0 +1,55 @@
...
+          just format
+          if [ $? -ne 0 ]; then
+            echo "::set-output name=format_failed::true"
+            exit 1
+          fi
+
+      - name: check for changes
+        id: check-changes
+        run: |
+          if [ -n "$(git status --porcelain)" ]; then
```
KendallWeihe, line 31: nice, TIL about the `--porcelain` parameter

.github/workflows/format.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: commit changes
+        if: steps.check-changes.outputs.changes_detected == "true"
+        run: |
+          git config --local user.email "github-actions[bot]@users.noreply.github.com"
+          git config --local user.name "github-actions[bot]"
+          git add "**/*.json"
+          git commit --message "chore: format json files"
+          git push
+
+      - name: create issue if formatting failed
```
KendallWeihe, line 44: oh this is fancy, nice!

Justfile
KendallWeihe: looks good!

KendallWeihe (approved):
💯

hosted/json-schemas/definitions.json
jiyoontbd: https://github.com/TBD54566975/tbdex/issues/371

PR #372, opened 2024-08-20
fix: simplify json format/validation action

lamchau:
Simplified the GitHub Actions workflow in PR #364 😅. This update also fixes an issue where `just` was missing in the `hermit` setup, which the previous workflow didn’t catch.

Since `just format` inherently includes `just lint`, we can just consolidate that into a single action.

This change removes much of the automation introduced in the earlier PR, placing the responsibility on the author to ensure well-formed JSON is committed to the repository. The original intent was to automate formatting when directly modifying JSON files via the GitHub web interface (bypassing the `pre-commit` hook). However, after reconsideration, this is likely not the most common workflow and might not be something we should encourage anyway.

Note: We still only check `hosted/*.json` which are the only assets published to tbdex.dev.


jiyoontbd (approved):

nitro-neal (approved):

Justfile
```diff
@@ -31,17 +31,15 @@ lint: clean
...
-        no_errors_found=false
+        has_errors=true
       fi
     done
 
-    if [ "$no_errors_found" = true ]; then
-      echo "[success] All JSON files are valid"
-    else
+    if [ "$has_errors" = true ]; then
       exit 1
```
KendallWeihe, line 43: maybe just `exit 1` right away instead of assigning to a variable and then reading afterwards?

KendallWeihe (approved):
nice, so each CI run should execute linting and if it fails then PRs can't be merged until the developer runs `just format` and commits the changes

Justfile
lamchau: we could! if the user had 3 malformed files then they'd only see one at a time.

this felt more "ergonomic" to show all the files so potentially it could be fixed without having to run it multiple times.

open to either approach!

jiyoontbd: i'm for showing all the files that fail lint at once

KendallWeihe: yeah nice, that's better!

